PR #2, opened 2019-01-08
Feature/enter name

dwybourn:
Added enter name functionality. Accessibility to me done in a future PR.


sample.env
```diff
@@ -1,3 +1,4 @@
 NODE_ENV=development
 PORT=8080
 SESSION_SECRET=xxxx-xxxx-xxxx-xxxx-xxxx
+CLAIM_BASE_URL=http://localhost:8090
```
mikeSmith08: Can we rename this to CLAIMANT_SERVICE_URL for clarity, took me a bit to figure out what this was really for.

src/web/routes/application/personal-details/sanitize.js
```diff
@@ -0,0 +1,18 @@
+const trimWhiteSpace = (body) => {
```
gethyn1: I think we can write this in a more functional manner

PR #5, opened 2019-01-09
Add smoke tests back into new UI project

mikeSmith08:


.cfignore
```diff
@@ -37,3 +37,5 @@ test/screenshots
 # ci build files
 bin
 build
+
+node_modules
```
YetAnotherMatt: The app is failing to deploy in CF - this is the most likely difference I can see between this and the hof version.

PR #10, opened 2019-01-10
Feature/enter name

dwybourn:
Added accessibility tests.


src/test/a11y/test-suite.js
```diff
@@ -1 +1,41 @@
...
+'use strict'
+
+const { getSIDCookieAndCSRFToken, postFormData } = require('./request')
+const pa11yWithSettings = require('./pally')
+const handleTestResults = require('./results')
+const IGNORE_RULES = require('./ignore-rules')
+
+const ENTER_NAME_URL = 'http://localhost:8080/enter-name'
+const CONFIRM_URL = 'http://localhost:8080/confirm'
+const COMPLETE_URL = 'http://localhost:8080/complete'
```
YetAnotherMatt: We shouldn't be hard-coding host or port

dwybourn: Updated to use PORT variable.

PR #11, opened 2019-01-10
HTBHF-249 Changed APP_PORT env variable to PORT

YetAnotherMatt:


.travis.yml
```diff
@@ -39,7 +39,7 @@ env:
   global:
     - APP_PATH=.
     - APP_NAME=help-to-buy-healthy-foods
-    - APP_PORT=8080
+    - PORT=8080
```
mikeSmith08: Wasn't this what it was originally? If so, why the change back again?

PR #13, opened 2019-01-10
HTBHFB-119 First pass at using Cucumber feature files

mikeSmith08:
I'm just putting up the PR to get some opinion on what I've done so far. The idea was to see if we could get Cucumber working with JS and it does look like its do-able, but before I do all the ATs for the Enter Name story, I'd like everyone's opinion please. We're all going to be doing plenty of these so if each of you could let me know if you're happy with this approach or not, and any pros/cons, that would be appreciated.

I'm not going to merge this PR and I know the test:acceptance target isn't working, but I'll fix that. Its also quite rough around the edges as I haven't done too much tidying, but I'm looking for your opinion as to whether this seems a good way forward - especially Gethyn. If you want to run it, use npm run test:cucumber with the app and Redis running. Thanks!


src/test/acceptance/features/steps.js
```diff
@@ -0,0 +1,67 @@
...
+  driver = driverManager.initialise()
+  overview = new Overview(driver)
+  enterName = new EnterName(driver)
+})
+
+After(function () {
+  driver.quit()
+})
+
+Given('I navigate to the HTBHF website and start the application process', async function () {
```
dwybourn: We should be starting on the enter-name url as the start page is something that we will not creating or maintaining.

dwybourn:
Looks good so far besides that one comment I mentioned.

src/test/acceptance/features/steps.js
mikeSmith08: Yep, good point, I'll correct that.

package.json
```diff
@@ -14,6 +14,7 @@
     "test:lint": "standard --env mocha",
     "test:unit": "nyc tape 'src/web/**/*.test.js'",
     "test:acceptance": "mocha 'src/test/acceptance/**/*.js'",
+    "test:cucumber": "./node_modules/.bin/cucumber-js src/test/acceptance/features/",
```
gethyn1: I think that `./node_modules/.bin/cucumber-js` could be replaced with `cucumber`

src/test/acceptance/features/enter_name.feature
```diff
@@ -0,0 +1,10 @@
+Feature:
```
gethyn1: Does the filename format need to be `enter_name.feature` rather than `enter-name.feature`?

gethyn1:
I'm not totally sure I understand how this works so would be good to go over in person .. but I think it looks good!!

PR #22, opened 2019-01-11
HTBHF-251 Claimant url is now part of the manifest file.

dwybourn:


application_dependencies.txt
```diff
@@ -0,0 +1 @@
+htbhf-claimant-service((suffix))
```
YetAnotherMatt: This file should be deleted

PR #39, opened 2019-01-16
Feature/enter nino route

dwybourn:


src/web/views/enter-nino.njk
```diff
@@ -0,0 +1,38 @@
...
+  <p class="govuk-body">{{ formDescription }}</p>
+
+  {% if errors %}
+    {{ govukErrorSummary({
+      titleText: "There is a problem",
+      errorList: errors | toErrorList
+    }) }}
+  {% endif %}
+
+  {% from "input/macro.njk" import govukInput %}
```
gethyn1: This is a duplicate import

src/web/views/enter-nino.njk
```diff
@@ -0,0 +1,38 @@
...
+  {{ govukFieldset({
+    legend: {
+      text: heading,
+      classes: "govuk-fieldset__legend--xl",
+      isPageHeading: true
+    }
+  }) }}
+  <p class="govuk-body">{{ formDescription }}</p>
+
+  {% if errors %}
```
gethyn1: This errors check is reused across templates - should we refactor?

src/web/views/enter-nino.njk
dwybourn: Removed.

PR #47, opened 2019-01-18
Feature/acceptance tests

dwybourn:


src/test/common/page/enter-name.js
```diff
@@ -15,10 +15,20 @@ class EnterName extends Page {
     return this.findById('first-name')
   }
 
+  async getFirstNameValue () {
+    const firstNameText = await this.getFirstNameField()
```
mikeSmith08: Should be `firstNameField` not `firstNameText`.

src/test/common/page/enter-name.js
```diff
@@ -15,10 +15,20 @@ class EnterName extends Page {
...
+    return firstNameText.getAttribute('value')
+  }
+
   async getLastNameField () {
     return this.findById('last-name')
   }
 
+  async getLastNameValue () {
+    const lastNameText = await this.getLastNameField()
```
mikeSmith08: Should be `lastNameField` not `lastNameText`.

src/test/common/steps/enter-name-steps.js
```diff
@@ -54,6 +54,16 @@ Then('I am informed that the first name is too long', async function () {
   expect(errorMessage).to.be.equal('Enter a shorter first or given name')
 })
 
+Then(/^I see the first name I entered in the textbox$/, async function () {
```
mikeSmith08: As this is specifically checking for the invalid long name, I'd prefer this step to be something like `I see the invalid first name entered in the textbox`, otherwise I'd expect it to be a valid name, or at least parameterised. Same with the method below too.

PR #51, opened 2019-01-18
Added compatibility tests

YetAnotherMatt:
We can now run compatibility tests using browserstack from web-ui. Currently this is only triggered manually, the separate compatibility tests project remains the active suite of tests for now.

### DIFF
--- a/src/test/compatibility/compatibility-report.js
+++ b/src/test/compatibility/compatibility-report.js
@@ -1,0 +1,11 @@
+var reporter = require('cucumber-html-reporter')
+
+var options = {
+  theme: 'bootstrap',
+  jsonFile: 'build/reports/compatibility-report.json',
+  output: 'build/reports/compatibility-report.html',
+  reportSuiteAsScenarios: true,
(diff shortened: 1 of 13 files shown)


src/test/compatibility/compatibility-report.js
```diff
@@ -0,0 +1,11 @@
+var reporter = require('cucumber-html-reporter')
```
gethyn1: Can these two instances of `var` be updated to `const`?

src/test/compatibility/config/constants.js
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable no-process-env */
+'use strict'
+
+require('dotenv').config()
+
+const APP_BASE_URL = process.env.APP_BASE_URL
+
+module.exports.COMPATIBILITY_BASE_URL = `${APP_BASE_URL}`
```
gethyn1: Does this need to be coerced to a string? I'm guessing it's already a string?

PR #54, opened 2019-01-21
HTBHF-293 Added accessibility tests.

dwybourn:


src/test/a11y/test-suite.js
```diff
@@ -7,9 +7,10 @@ const handleTestResults = require('./results')
 const IGNORE_RULES = require('./ignore-rules')
 const { PORT } = require('../common/config')
 
+const ENTER_NINO_URL = `http://localhost:${PORT}/enter-nino`
```
mikeSmith08: Could we create a variable such as BASE_URL for `http://localhost:${PORT}` to remove duplication in these constants?

dwybourn: Done.

PR #59, opened 2019-01-22
HTBHF-250 i18n acceptance tests.

dwybourn:


src/test/common/page/page.js
```diff
@@ -23,9 +23,9 @@ class Page {
     this.driver = driver
   }
 
-  async open (url) {
+  async open (url, lang = 'en') {
     try {
-      return this.driver.get(url)
+      return this.driver.get(`${url}?lang=${lang}`)
```
YetAnotherMatt: Can we default to not specifying the lang query param?

src/test/acceptance/features/internationalisation.feature
```diff
@@ -0,0 +1,13 @@
+Feature:
+  In order apply for HTBHF programme
+  As a potential claimant
+  I want to view the site in my preferred language
+
+  Scenario:
+    When I go to the enter national insurance number page with Welsh language selected
+    Then The enter national insurance page is in Welsh
```
YetAnotherMatt: 'The' should be 'the'

dwybourn: Change to lower case 't'

PR #67, opened 2019-01-23
HTBHF-370 Now posting the date of birth to the claimant service.

dwybourn:


src/web/views/check.njk
```diff
@@ -54,6 +54,7 @@
   <input type="hidden" name="firstName" value="{{ claim.firstName }}" />
   <input type="hidden" name="lastName" value="{{ claim.lastName }}" />
   <input type="hidden" name="nino" value="{{ claim.nino }}" />
+  <input type="hidden" name="dateOfBirth" value="{{ claim.dateOfBirth }}" />
```
YetAnotherMatt: I don't think we should be relying on unvalidated input from the browser (these hidden fields) to determine what values we send to the claimant service

dwybourn: Removing hidden fields and now sending data from the session instead.

PR #74, opened 2019-01-24
HTBHF-9 Deleting language cookie at the end of of the apply journey.

dwybourn:


src/web/routes/application/confirm.js
```diff
@@ -4,6 +4,8 @@ const pageContent = {
 }
 
 const getConfirmPage = (req, res) => {
+  req.session.destroy()
```
gethyn1: @mikeSmith08 @YetAnotherMatt what are your thoughts on destroying the session on the confirm page? Currently we can hit this page directly so could in theory kill the session before POSTing the form.

mikeSmith08: My opinion on this is that as we don't provide navigation direct to this via any means apart from the user messing with the URL, then its fine that this happens - I'd expect the final page of the flow to so this.

PR #85, opened 2019-01-25
HTBHF-10 Added compatibility tests to 'npm test'

YetAnotherMatt:


ci_scripts/ci_deploy.sh
```diff
@@ -34,6 +34,7 @@ rm -rf DepartmentOfHealth-htbhf-htbhf-deployment-scripts-*
 rm deployment-scripts.zip
 
 export SCRIPT_DIR=${BIN_DIR}/deployment-scripts
+export APP_VERSION=$(sed -nE 's/^[ \\t]*"version": "([0-9]{1,}\.[0-9]{1,}\.[0-9x]{1,})",$/\1/p' package.json;)
```
YetAnotherMatt: This adds APP_VERSION so the deployment scripts can create an environment variable when deploying to CF with the version deployed

PR #86, opened 2019-01-25
HTBHF-402 Added acceptance tests for are you pregnant. 

dwybourn:


src/test/acceptance/features/are-you-pregnant.feature
```diff
@@ -0,0 +1,15 @@
...
+  I want to select whether or not I am pregnant
+
+  Background:
+    Given I am on the are you pregnant page
+
+  Scenario: Select that I am pregnant
+    When I select that I am pregnant
+    Then I am shown the check details page
+
+  Scenario: No not select an option
```
YetAnotherMatt: "No not" -> "Do not"

src/test/common/page/enter-details-page.js
```diff
@@ -0,0 +1,14 @@
+const Page = require('./page')
+
+class EnterDetailsPage extends Page {
```
YetAnotherMatt: Maybe call this class something like 'DataEntryPage'?

src/test/common/page/are-you-pregnant.js
```diff
@@ -1,20 +1,30 @@
 'use strict'
 
-const Page = require('./page')
+const EnterDetailsPage = require('./enter-details-page')
 const OVERVIEW_PAGE_TITLE = 'GOV.UK - Are you pregnant?'
```
mikeSmith08: Shouldn't be called OVERVIEW_PAGE_TITLE, should be PAGE_TITLE or similar.

src/test/acceptance/features/are-you-pregnant.feature
```diff
@@ -0,0 +1,26 @@
...
+  Background:
+    Given I am on the are you pregnant page
+
+  Scenario: No option is select
+    Then No option is selected
+
+  Scenario: Yes and No are displayed
+    Then Yes and No options are displayed
+
+  Scenario Outline: Select that an option
```
YetAnotherMatt: "Select that an option" -> "Select an option"
Worth bearing in mind that in the next story we'll be changing what happens when selecting 'yes', so this will change anyway at that point. We can probably leave this for now.

PR #90, opened 2019-01-26
Feature/compatibility tests

YetAnotherMatt:
Replaced Gruntfile with npm-run-all and separated the json output into multiple files with with browser metadata.
Enabled cucumber reporting for multiple json files.


package.json
```diff
@@ -19,19 +19,32 @@
...
+    "test:compatibility:local:report": "node src/test/compatibility/local-compatibility-report.js",
     "test:accessibility": "node 'src/test/a11y/test-suite.js'",
     "test:browser": "npm run test:wait && npm-run-all -s test:smoke test:accessibility test:compatibility:local test:acceptance test:acceptance:report",
     "test:wait": "./npm_scripts/wait_for_app_to_start.sh",
     "coverage": "nyc report --reporter=lcov --reporter=text-lcov > coverage.lcov && codecov",
     "build": "npm run build:sass",
     "build:sass": "node-sass node_modules/govuk-frontend/all.scss src/web/assets/styles/all.css",
-    "postinstall": "npm run build"
+    "postinstall": "npm run build",
+    "test:compatibility": "mkdir -p build/reports/compatibility && cross-env TESTS=compatibility npm-run-all -c -p --max-parallel 1 compat:win compat:osx compat:mobile",
```
YetAnotherMatt: The compatibility tests are divided into 3 groups: compat:win compat:osx compat:mobile
'--max-parallel 1' ensures only one group is run at a time, '-c' ensures all groups are run, even if one reports an error (npm-run-all will still report an error if any test failed)

package.json
```diff
@@ -19,19 +19,32 @@
...
+    "test:compatibility:debug": "node $NODE_DEBUG_OPTION node_modules/cucumber/bin/cucumber-js src/test/compatibility/features/ --require 'src/test/common/steps/*.js' --require 'src/test/compatibility/config' --format ./src/test/compatibility/custom-json-formatter.js:build/reports/compatibility-report.json",
+    "test:compatibility:report": "node src/test/compatibility/compatibility-report.js",
+    "compat:win": "cross-env BROWSER_STACK_OS=Windows BROWSER_STACK_OS_VERSION=10 npm-run-all -c -p --max-parallel 5 compat:win:ie compat:win:edge compat:win:chrome compat:win:firefox",
```
YetAnotherMatt: within each group of compatibility tests, tests run in parallel (up to 5 at a time, as this is the most that can run in parallel on BrowserStack)

package.json
```diff
@@ -19,19 +19,32 @@
...
-    "test:compatibility:debug": "cross-env TESTS=compatibility node $NODE_DEBUG_OPTION node_modules/cucumber/bin/cucumber-js src/test/compatibility/features/ --require 'src/test/common/steps/*.js' --require 'src/test/compatibility/config' -f json:build/reports/compatibility-report.json",
-    "test:compatibility:report": "node src/test/compatibility/compatibility-report.js",
     "test:compatibility:local": "npm-run-all -s test:compatibility:local:test test:compatibility:local:report",
-    "test:compatibility:local:test": "mkdir -p build/reports && cross-env TESTS=localCompatibility cucumber-js src/test/compatibility/features/ --require 'src/test/common/steps/*.js' --require 'src/test/compatibility/config' -f json:build/reports/compatibility-report.json",
-    "test:compatibility:local:report": "node src/test/compatibility/compatibility-report.js",
+    "test:compatibility:local:test": "mkdir -p build/reports/compatibility && cross-env TESTS=localCompatibility cucumber-js src/test/compatibility/features/ --require 'src/test/common/steps/*.js' --require 'src/test/compatibility/config' -f json:build/reports/compatibility-report.json",
```
mikeSmith08: You should add the debug settings back into this one presuming this is the one that we can use to debug/ run locally.

package.json
```diff
@@ -19,19 +19,32 @@
...
     "test:accessibility": "node 'src/test/a11y/test-suite.js'",
     "test:browser": "npm run test:wait && npm-run-all -s test:smoke test:accessibility test:compatibility:local test:acceptance test:acceptance:report",
     "test:wait": "./npm_scripts/wait_for_app_to_start.sh",
     "coverage": "nyc report --reporter=lcov --reporter=text-lcov > coverage.lcov && codecov",
     "build": "npm run build:sass",
     "build:sass": "node-sass node_modules/govuk-frontend/all.scss src/web/assets/styles/all.css",
-    "postinstall": "npm run build"
+    "postinstall": "npm run build",
+    "test:compatibility": "mkdir -p build/reports/compatibility && cross-env TESTS=compatibility npm-run-all -c -p --max-parallel 1 compat:win compat:osx compat:mobile",
+    "test:compatibility:debug": "node $NODE_DEBUG_OPTION node_modules/cucumber/bin/cucumber-js src/test/compatibility/features/ --require 'src/test/common/steps/*.js' --require 'src/test/compatibility/config' --format ./src/test/compatibility/custom-json-formatter.js:build/reports/compatibility-report.json",
```
mikeSmith08: Ok, now I'm confused by all the targets in here, how is this one different to local:test? Think this means at least we need to do some renaming of targets.

package.json
YetAnotherMatt: This is the one that runs as part of`'npm:test`

YetAnotherMatt:
Closing the PR until we have a better way to invoke the tests

PR #92, opened 2019-01-28
HTBHF-412 Add conditional EDD field to UI

gethyn1:
- Add conditional EDD field to UI
- Refactor date input field into common nunjucks macro

### DIFF
--- a/src/web/server/locales/en/common.json
+++ b/src/web/server/locales/en/common.json
@@ -26,7 +26,9 @@
   },
   "areYouPregnant": {
     "title": "Are you pregnant?",
-    "heading": "Are you pregnant?"
+    "heading": "Are you pregnant?",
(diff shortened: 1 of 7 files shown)


src/web/server/locales/en/common.json
```diff
@@ -26,7 +26,9 @@
   },
   "areYouPregnant": {
     "title": "Are you pregnant?",
-    "heading": "Are you pregnant?"
+    "heading": "Are you pregnant?",
+    "expectedDateOfDeliveryText": "When is your baby due to be born?",
+    "expectedDateOfDeliveryHint": "For example, 31 3 2019"
```
YetAnotherMatt: Weren't we going to be making this always 6 months in the future?

gethyn1: Yes .. There is another sub-task for this.

PR #95, opened 2019-01-29
Feature/validate expected delivery date

gethyn1:
- Added validation in range 1 month in past -> 8 months in future
- Fix acceptance tests for happy path only
- Refactor date formatters

### DIFF
--- a/src/web/routes/application/are-you-pregnant/validate.js
+++ b/src/web/routes/application/are-you-pregnant/validate.js
@@ -1,10 +1,13 @@
 const { check } = require('express-validator/check')
+const { isNil } = require('ramda')
 const { toDateString } = require('../common/formatters')
 const { isValidDate, isDateMoreThanOneMonthAgo, isDateMoreThanEightMonthsInTheFuture } = require('../common/validators')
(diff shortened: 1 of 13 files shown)


src/web/routes/application/are-you-pregnant/validate.js
```diff
@@ -1,9 +1,49 @@
 const { check } = require('express-validator/check')
+const { toDateString } = require('../common/formatters')
+const { isValidDate, isDateMoreThanOneMonthAgo, isDateMoreThanEightMonthsInTheFuture } = require('../common/validators')
+const { YES, NO } = require('./constants')
+
+const addDateToBody = (req, res, next) => {
+  if (req.body.areYouPregnant === NO) {
```
dwybourn: Can we have a test case for this scenario please. I'd also rename this method and extract the part that adds the Date to the body.

gethyn1: Added unit tests but don't agree with "extract the part that adds the Date to the body" as this is the main purpose of the function.

PR #96, opened 2019-01-29
HTBHF-439 Fixed example due date to always be in the future

YetAnotherMatt:


src/web/routes/application/are-you-pregnant/are-you-pregnant.js
```diff
@@ -19,6 +19,14 @@ const areYouPregnant = {
   validate
 }
 
+const exampleDate = (fromDate = new Date()) => {
+  let future = new Date(fromDate)
```
gethyn1: `future` never gets reassigned so should probably be a `const`

src/web/routes/application/are-you-pregnant/are-you-pregnant.js
```diff
@@ -8,7 +8,7 @@ const pageContent = ({ translate }) => ({
   yes: translate('yes'),
   no: translate('no'),
   expectedDeliveryDateText: translate('areYouPregnant.expectedDeliveryDateText'),
-  expectedDeliveryDateHint: translate('areYouPregnant.expectedDeliveryDateHint')
+  expectedDeliveryDateHint: translate('areYouPregnant.expectedDeliveryDateHint', { exampleDate: exampleDate() })
```
gethyn1: It's probably best to define `exampleDate()` before its used. In some situations this could lead to a reference error, see https://eslint.org/docs/rules/no-use-before-define

PR #99, opened 2019-01-29
Feature/validation messages

gethyn1:
Refactor function for translating validation messages using express validator to avoid redefining the function signature on each validation

### DIFF
--- a/src/web/routes/application/common/translate-validation-message.js
+++ b/src/web/routes/application/common/translate-validation-message.js
@@ -1,0 +1,13 @@
+/**
+ * `translateValidationMessage()` follows the express-validator signature for
+ * defining custom validation messages. `req.t()` is the internationalisation
+ * function provided by i18next-express-middleware.
+ *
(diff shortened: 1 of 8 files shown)


src/web/routes/application/common/translate-validation-message.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * `translateValidationMessage()` follows the express-validator signature for
+ * defining custom validation messages. `req.t()` is the internationalisation
+ * function provided by i18next-express-middleware.
+ *
+ * - https://express-validator.github.io/docs/custom-error-messages.html
+ * - https://github.com/i18next/i18next-express-middleware
+ */
+const translateValidationMessage = message => (value, { req }) => req.t(message)
```
gethyn1: Previously the validation translation fn looked like:
```
(value, { req }) => req.t(message, { value })
```
I have removed `{ value }` as it didn't seem to be used.

PR #109, opened 2019-01-31
HTBHF-452 Added a transformData process to correctly save the desired data to the session

YetAnotherMatt:


src/web/routes/application/are-you-pregnant/transform-data.js
```diff
@@ -0,0 +1,16 @@
...
+      req.body['expectedDeliveryDate-day'],
+      req.body['expectedDeliveryDate-month'],
+      req.body['expectedDeliveryDate-year']
+    )
+    return { areYouPregnant: req.body.areYouPregnant, expectedDeliveryDate: expectedDeliveryDate }
+  }
+  return { areYouPregnant: req.body.areYouPregnant, expectedDeliveryDate: null }
+}
+
+module.exports.transformData = transformData
```
gethyn1: Just as a point of interest, this file could be written like this:
```
const transformData = (req) => {
  const { areYouPregnant } = req.body

  if (areYouPregnant === YES) {
    const expectedDeliveryDate = toDateString(
      req.body['expectedDeliveryDate-day'],
      req.body['expectedDeliveryDate-month'],
      req.body['expectedDeliveryDate-year']
    )
    return {
      areYouPregnant,
      expectedDeliveryDate
    }
  }
  
  return {
    areYouPregnant,
    expectedDeliveryDate: null
  }
}
```
obviously very subjective, but as a JS dev I find that more readable (I know I'm not supposed to make subjective readability comments on PRs!!).

src/web/routes/application/common/populate-session.test.js
```diff
@@ -0,0 +1,30 @@
+const test = require('tape')
+const sinon = require('sinon')
+const { populateSession } = require('./populate-session')
+
+test('populateSession() should copy fields from the request body by default', (t) => {
```
gethyn1: Should this be testing `defaultTransform()` directly, and the `populateSession()` test just assert that the default function is called?

src/web/routes/application/are-you-pregnant/transform-data.js
YetAnotherMatt: your version does look better

PR #112, opened 2019-02-01
HTBHF-452 ensure expectedDeliveryDate is null when not pregnant

YetAnotherMatt:


gethyn1:
I would separate the transformer logic into a new file to keep those files small and easy to mentally parse - but just a suggestion so approving anyway.

PR #113, opened 2019-02-01
HTBHF-458 Added assertions for the text on validation error links.

dwybourn:


YetAnotherMatt:
While you're looking at the relevant area, how about adding an assertion that the field in error still contains the invalid value? I ask because while I managed to remove all the persisted values when the page had errors, and none of the tests failed.

PR #115, opened 2019-02-01
Feature/check details - update the check details page to match the prototype.

mikeSmith08:


src/web/views/check.njk
```diff
@@ -1,85 +1,74 @@
...
         [
           {
-            text: "Are you pregnant?"
+            text: areYouPregnant
           },
           {
             text: claim.areYouPregnant
           }
         ],
+        expectedDeliveryDateRow,
```
gethyn1: I'm unsure about having a single row that is built on the server - would it be better to build the complete `rows` array on the server?

The `pageContent()` function now seems like its performing two functions:
1. providing text strings to the template
2. creating structure for rendering HTML

In my mind these are two distinct functions.

src/web/routes/application/check/get.js
```diff
@@ -1,12 +1,38 @@
...
+  heading: translate('check.heading'),
+  name: translate('check.name'),
+  nino: translate('check.nationalInsuranceNumber'),
+  dateOfBirth: translate('check.dateOfBirth'),
+  areYouPregnant: translate('check.areYouPregnant'),
+  dueDate: translate('check.dueDate'),
+  address: translate('check.address'),
+  sendApplicationHeader: translate('check.sendApplicationHeader'),
+  sendApplicationText: translate('check.sendApplicationText'),
+  buttonText: translate('check.buttonText'),
```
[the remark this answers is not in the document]
mikeSmith08: Ok, didn't know about this, but have done it now.

src/web/views/check.njk
mikeSmith08: Ok, I'll build them all on the server, I'll work on it now, might be Monday before I push that commit.

PR #118, opened 2019-02-04
Feature/check details

mikeSmith08:
- Update check details page in line with prototype
- Move all building of row data for check details page into JavaScript so we can dynamically build this data based on the claimant's answers
- Make sure that Baby's due date is only shown if Are you pregnant answer is Yes.
- Capitalise the Are you pregnant answer
- Make sure that each line of the address is delimited with <br/>
- Make sure that optional address lines are removed if they are not entered
- Make sure all text is translated
- Unit test all code building up dynamic row data.


src/web/routes/application/check/get.js
```diff
@@ -1,16 +1,109 @@
...
+  return buildRowData(
+    translate('check.dateOfBirth'),
+    [claim['dateOfBirth-day'], claim['dateOfBirth-month'], claim['dateOfBirth-year']].join(' ')
+  )
+}
+
+const buildAreYouPregnantRow = (translate, claim) => {
+  return buildRowData(
+    translate('check.areYouPregnant'),
+    capitaliseString(claim.areYouPregnant)
```
YetAnotherMatt: This raises an interesting point about translations - we're not translating yes/no. Could we instead look up a translation for 'yes' and 'no'? Probably one for another pr...

src/web/routes/application/check/get.js
```diff
@@ -1,16 +1,109 @@
...
+const buildExpectedDeliveryDateRow = (translate, claim) => {
+  if (claim.areYouPregnant === YES) {
+    return buildRowData(
+      translate('check.dueDate'),
+      [claim['expectedDeliveryDate-day'], claim['expectedDeliveryDate-month'], claim['expectedDeliveryDate-year']].join(' ')
+    )
+  }
+}
+
+// We need to build up this row here because we cannot dynamically hide a using the govukTable component based on session data.
```
YetAnotherMatt: I think this comment can probably go now, as it applies to all rows

YetAnotherMatt: Raised as a subtask: HTBHF-472

src/web/routes/application/check/get.js
YetAnotherMatt: Raised as a subtask: HTBHF-472

mikeSmith08: Yep, you're right, the latest commit translates it.

src/web/routes/application/check/get.js
mikeSmith08: Removed.

src/web/routes/application/check/get.js
```diff
@@ -1,16 +1,97 @@
...
-  heading: 'Check'
+const { YES } = require('../are-you-pregnant/constants')
+
+const pageContent = ({ translate, claim }) => ({
+  title: translate('check.title'),
+  heading: translate('check.heading'),
+  sendApplicationHeader: translate('check.sendApplicationHeader'),
+  sendApplicationText: translate('check.sendApplicationText'),
+  buttonText: translate('buttons:acceptAndSend'),
+  checkRowData: buildCheckRowData(translate, claim)
```
gethyn1: Subjective suggestion:

I would maybe pass `checkRowData` directly to the template (in `render` call). This would avoid passing `claim` to `pageContent()` just to pass it to `buildCheckRowData()`.

The `checkRowData` data structure is also quite different to the translated strings so maybe makes sense to separate.

src/web/routes/application/check/get.js
```diff
@@ -1,16 +1,97 @@
...
+
+const buildCheckRowData = (translate, claim) => {
+  return [
+    buildNameRow(translate, claim),
+    buildNinoRow(translate, claim),
+    buildDateOfBirthRow(translate, claim),
+    buildAreYouPregnantRow(translate, claim),
+    buildExpectedDeliveryDateRow(translate, claim),
+    buildAddressRow(translate, claim)
+  ].filter(row => row !== undefined)
```
gethyn1: Could make use of Ramda `isNil` or javascript `Boolean` here:
```
const buildCheckRowData = (translate, claim) => [
  buildNameRow(translate, claim),
  buildNinoRow(translate, claim),
  buildDateOfBirthRow(translate, claim),
  buildAreYouPregnantRow(translate, claim),
  buildExpectedDeliveryDateRow(translate, claim),
  buildAddressRow(translate, claim)
].filter(!isNil)
```

src/web/routes/application/check/get.js
```diff
@@ -1,16 +1,97 @@
...
+}
+
+const buildNameRow = (translate, claim) => {
+  return buildRowData(
```
gethyn1: Suggestion: Could make use of arrow function implicit returns on a number of these functions (not a PR blocker):
```
const buildNameRow = (translate, claim) => buildRowData(
  translate('check.name'),
  [claim.firstName, claim.lastName].join(' ')
)
```

src/web/views/check.njk
```diff
@@ -1,85 +1,31 @@
 {% extends "templates/form.njk" %}
 {% from "input/macro.njk" import govukInput %}
 {% from "table/macro.njk" import govukTable %}
+{% from "fieldset/macro.njk" import govukFieldset %}
 
-{% set buttonText = "Confirm" %}
+{% set safeRows = [] %}
+{% for heading, content in checkRowData %}
+  {% set row = [{ text: heading.text }, { text: content.text | safe } ] %}
```
gethyn1: Is there any situation where the `| safe` filter here could be a security vulnerability?

src/web/routes/application/check/get.js
mikeSmith08: Done

src/web/routes/application/check/get.js
mikeSmith08: Done.

src/web/routes/application/check/get.js
mikeSmith08: Done.

src/web/routes/application/check/get.js
```diff
@@ -1,16 +1,97 @@
...
+
+const buildAddressRow = (translate, claim) => {
+  return buildRowData(
+    translate('check.address'),
+    [
+      claim.addressLine1,
+      claim.addressLine2,
+      claim.townOrCity,
+      claim.postcode
+    ].filter(item => (item !== undefined && item.length !== 0))
```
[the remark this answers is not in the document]
mikeSmith08: Done. I think your preferences pretty much have to define our coding standards at the moment as no one else has any great experience with JS, so happy to go with that.

src/web/views/check.njk
mikeSmith08: Possibly, but I think it should be filtered on the input sanitisation rather than the output here. If you have another solution which will work with Nunjucks then please let me know.

src/web/routes/application/check/get.test.js
```diff
@@ -0,0 +1,162 @@
...
+
+test('isNilOrEmpty', (t) => {
+  t.equal(isNilOrEmpty(''), true, 'should be marked as empty')
+  t.equal(isNilOrEmpty(null), true, 'should be marked as empty')
+  t.equal(isNilOrEmpty(undefined), true, 'should be marked as empty')
+  t.equal(isNilOrEmpty('Something'), false, 'should not be marked as empty')
+  t.end()
+})
+
+const testTranslate = (key) => {
```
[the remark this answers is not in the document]
mikeSmith08: Done

PR #127, opened 2019-02-07
Modify date format in the check details page to display month in full.

mikeSmith08:
This matches the GDS standards here: https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style (look under D for dates).


src/web/routes/application/common/formatters.js
```diff
@@ -21,7 +23,22 @@ const dateAsString = ({ date = new Date(), monthAdjustment = 0 } = {}) => {
...
 }
 
+// TODO - When we are no longer able to go directly to a page within the flow (which makes ATs simpler)
+// we should reinforce the fact that day, month and year should be mandatory.
+const buildFormattedDateForDisplay = (day, month, year) => {
+  return moment({ years: year, months: month - 1, date: day }).format(DATE_FORMAT)
+}
+
+const buildFormattedDateForDisplayFromDate = (date) => {
+  if (date === undefined || !(date instanceof Date)) {
```
gethyn1: Just something to be aware of; I would generally use the check `typeof date === 'undefined'` as `date === undefined` in certain situations can return `"date is not defined"` .. in this situation however I think its fine as `date` has definitely been declared as a parameter.

PR #128, opened 2019-02-08
 HTBHF-489 added a state machine to control the flow of the application

YetAnotherMatt:


src/web/routes/application/constants.js
```diff
@@ -1,2 +1,3 @@
 module.exports.YES = 'yes'
 module.exports.NO = 'no'
+module.exports.CHECK_URL = '/check'
```
gethyn1: Not to do as part of this PR, but I think the `application` directory needs slightly restructuring; a lot of the files in `application/common` are not actually common and could live in `application/middleware`. This would leave `common` for actual common files such as this `constants.js`:

```
- /application
|- /common
|  |- constants.js
|
| - /middleware
|  |- /render-view
| ...
```

src/web/routes/application/check/get.js
```diff
@@ -27,6 +28,8 @@ const getCheck = (req, res) => {
   const stepArrays = steps.map(getRowData(req))
   const checkRowData = flatten(stepArrays)
 
+  stateMachine.setState(states.IN_REVIEW, req)
```
gethyn1: I think this is the wrong place for updating state, but again can be handled in another PR

PR #131, opened 2019-02-08
HTBHF-481 added an acceptance test for edit your answers

YetAnotherMatt:


src/test/common/page/page.js
```diff
@@ -73,6 +73,15 @@ class Page {
     }
   }
 
+  async findByXPath (xpath) {
+    try {
```
dwybourn: All the other find by methods wait for the element first before finding it. I'm guessing this isn't need?

YetAnotherMatt: This isn't needed, I think because where we use it the page is already loaded. I also couldn't figure out how to make that part work with xpath.

src/test/common/steps/check-details.js
```diff
@@ -141,3 +147,8 @@ function assertAddressShown (tableContents, expectedAddress) {
   const addressValue = getValueForField(tableContents, 'Address')
   expect(addressValue).to.be.equal(expectedAddress)
 }
+
+function assertHeaderAndChangeLinkShownForEachRow (tableContents) {
+  const matchingRows = tableContents.filter((value) => value.action.text === 'Change' && value.header.trim().length > 0)
+  expect(matchingRows.length).to.be.equal(tableContents.length)
```
mikeSmith08: You can use `expect().to.have.lengthOf()` for this assertion.

PR #135, opened 2019-02-11
HTBHF-275 add security middleware

gethyn1:
- Add Helmet to secure express server
- Add content security policy


dwybourn:
Should we also add something like https://github.com/hengkiardo/express-enforces-ssl to make sure the app will only serve on https?

PR #136, opened 2019-02-11
HTBHF-498 Add extra sanitisation to the middleware such that HTML cha…

mikeSmith08:
…racters are escaped properly.

### DIFF
--- a/src/web/routes/application/middleware/sanitize.js
+++ b/src/web/routes/application/middleware/sanitize.js
@@ -1,8 +1,6 @@
 const { map, trim, compose } = require('ramda')
 const escapeHtml = require('escape-html')
 
-const trimValues = map(trim)
-
 const sanitiseInput = compose(escapeHtml, trim)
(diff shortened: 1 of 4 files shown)


src/web/routes/application/middleware/sanitize.js
```diff
@@ -1,13 +1,16 @@
 const { map, trim } = require('ramda')
+const escapeHtml = require('escape-html')
 
 const trimValues = map(trim)
+const escapeValues = map(escapeHtml)
 
 const sanitize = (req, res, next) => {
-  req.body = trimValues(req.body)
+  req.body = escapeValues(trimValues(req.body))
```
gethyn1: I think this line is mapping over all input values twice. A more efficient way to write this would be something like:
```
const sanitiseInput = compose(escapeHtml, trim)
const sanitiseBody = map(sanitiseBody)
req.body = sanitiseBody(req.body)
```

PR #138, opened 2019-02-12
HTBHF-497 Add acceptance test with script tags in it to test sanitising.

mikeSmith08:


src/test/acceptance/features/check-details.feature
```diff
@@ -29,3 +29,8 @@ Feature: Check details
     And I click continue
     Then I am shown the check details page
     And the check details page contains all data entered for a woman who is not pregnant
+
+  Scenario: An application for a woman who is not pregnant contains malicious input
```
YetAnotherMatt: Very minor, but I'd suggest that the fact that the applicant is not pregnant isn't relevant to the scenario. How about 'An application contains malicious input'?

PR #144, opened 2019-02-12
NOJIRA Increased browserstack timeout to 2 minutes

YetAnotherMatt:


YetAnotherMatt:
@mikeSmith08 it might be worth merging this first thing, so we can (hopefully) get a green build again

PR #145, opened 2019-02-14
Feature/submit page

mikeSmith08:
Update submit page to match content and make it translatable. The choice was made to include the language based content as a complete HTML rather than a large number of individual substitutions (paragraph1, paragraph2 etc) which would be come a maintenance nightmare.


src/web/routes/application/confirm.js
```diff
@@ -1,12 +1,22 @@
-const pageContent = {
-  title: 'Application successful',
-  heading: 'Application successful'
+const { isNilOrEmpty } = require('./common/predicates')
+
+const getLanguageBase = (language) => {
+  if (isNilOrEmpty(language)) {
```
YetAnotherMatt: I don' think we should assume that we always get a language: https://stackoverflow.com/a/41738594

src/web/views/locale/en/confirm.html
```diff
@@ -0,0 +1,16 @@
...
+<p class="govuk-body">We will send your prepaid card to the address you gave us.</p>
+<h3 class="govuk-heading-m">2. Money will be added to your card</h3>
+<p class="govuk-body">We will add money to your card every 4 weeks so that you can buy healthy food and milk. You will get £12.40 every 4 weeks until your baby is born.</p>
+<h3 class="govuk-heading-m">3. You can use your card to pay for fruit, vegetables and milk</h3>
+<p class="govuk-body">You can find out <a class="govuk-link" href="related/how-to-buy-with-prepaid-card">how to use your prepaid card and what you can buy with your money on GOV.UK (link opens in a new tab)</a></p>
+<div class="govuk-warning-text">
+    <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+    <strong class="govuk-warning-text__text">
+        <span class="govuk-warning-text__assistive">Warning</span>
+        When your baby is born you must telephone XXX to tell us. We will then increase your money to £6.20 each week until your baby is 1 year old.
```
YetAnotherMatt: I think Brian has a support telephone now - he should be able to provide the number

src/web/routes/application/confirm.js
mikeSmith08: Yes, you are right, the language header is optional in the request, but the package we use to deal with languages has a default language set (in `internationalisation.js`) of 'en' so if none is provided, then that default will be used, so by the time we get to the confirm page, we will have a language in the request object.

src/web/views/locale/en/confirm.html
mikeSmith08: The comment from Brian is that we should leave it this way for now as we don't know what this should be and hence should leave it like this so that its obvious we need to revisit this page.

PR #147, opened 2019-02-14
Feature/submit acceptance tests

mikeSmith08:
- Add acceptance test for confirm page
 - Modify compatibility tests to verify both the check and confirm pages
 - Refactor out wiremock code from a11y so it can be used by ATs.

One point of discussion is where the deletion of the Wiremock stubs currently live - I was initially going to add it to the `After` cucumber hook but thought this would be too expensive for all tests so I decided to add it to the final steps of tests which setup Wiremock mocks. The downside of this is that we need to remember to delete the mocks once we have set them up rather than them be automatically deleted. Happy to discuss...


src/test/common/wiremock.js
```diff
@@ -0,0 +1,26 @@
+const { postJsonData, performDelete } = require('./request')
+
+const WIREMOCK_MAPPING_URL = process.env.WIREMOCK_URL ? `${process.env.WIREMOCK_URL}/__admin/mappings` : 'http://localhost:8090/__admin/mappings'
+
+const SUCCESSFUL_CLAIMS_MAPPING = `{
```
gethyn1: You could maybe use `JSON.stringify()` here - a template string feels a bit odd, but maybe its OK?

PR #151, opened 2019-02-15
HTBHF-536 send request ID to claimant service

gethyn1:


src/web/routes/application/check/post.js
```diff
@@ -38,6 +38,9 @@ const postCheck = (config) => async (req, res, next) => {
     await request.post({
       uri: `${config.environment.CLAIMANT_SERVICE_URL}${CLAIMS_ENDPOINT}`,
       json: true,
+      headers: {
+        'X-Request-ID': req.headers['X-Request-ID']
```
YetAnotherMatt: Can we add 'X-SESSION-ID' as well?

PR #153, opened 2019-02-15
HTBHF-534 Fix setup (or not) of Wiremock stubs for compatibility tests.

mikeSmith08:


src/test/common/steps/common-steps.js
```diff
@@ -93,12 +95,24 @@ async function completeTheApplicationAsAWomanWhoIsNotPregnant () {
   await enterCardAddress(ADDRESS_LINE_1, ADDRESS_LINE_2, TOWN, POSTCODE)
 }
 
+async function setupWiremockMappings () {
```
gethyn1: I would maybe have these in a new separate file as they are used in multiple places

PR #159, opened 2019-02-18
HTBHF-534 Make sure we use the latest browser versions for compatibility tests at BrowserStack.

mikeSmith08:


package.json
```diff
@@ -36,14 +36,14 @@
...
     "compat:win": "cross-env BROWSER_STACK_OS=Windows BROWSER_STACK_OS_VERSION=10 npm-run-all -c -p --max-parallel 5 compat:win:ie compat:win:edge compat:win:chrome compat:win:firefox",
-    "compat:win:ie": "cross-env BROWSER_STACK_BROWSER=IE BROWSER_STACK_BROWSER_VERSION=11.0 cucumber-js src/test/compatibility/features/ --require 'src/test/common/steps/*.js' --require 'src/test/compatibility/config' --format ./src/test/compatibility/custom-json-formatter.js:build/reports/compatibility/win-ie-report.json",
+    "compat:win:ie": "cross-env BROWSER_STACK_BROWSER=IE cucumber-js src/test/compatibility/features/ --require 'src/test/common/steps/*.js' --require 'src/test/compatibility/config' --format ./src/test/compatibility/custom-json-formatter.js:build/reports/compatibility/win-ie-report.json",
     "compat:win:edge": "cross-env BROWSER_STACK_BROWSER=Edge BROWSER_STACK_BROWSER_VERSION=18.0 cucumber-js src/test/compatibility/features/ --require 'src/test/common/steps/*.js' --require 'src/test/compatibility/config' --format ./src/test/compatibility/custom-json-formatter.js:build/reports/compatibility/win-edge-report.json",
```
mikeSmith08: There is currently a bug with BrowserStack where the latest version of Edge actually points to version 16.0 rather than 18.0 which I have raised with them, so I'll change this when that is fixed. In the meantime it was actually handy to make sure that works when we specify a mix of latest versions of browsers and specific versions.

src/test/compatibility/custom-json-formatter.js
```diff
@@ -1,14 +1,30 @@
 const { JsonFormatter } = require('cucumber')
+const { notIsNilOrEmpty } = require('../../web/routes/application/common/predicates')
 
 require('dotenv').config()
 
+function isMobileDevice () {
+  return !!process.env.BROWSER_STACK_DEVICE
+}
+
+function getBrowserVersion () {
```
mikeSmith08: @gethyn1 there is almost certainly a nicer version of this, any good suggestions?

src/test/compatibility/custom-json-formatter.js
```diff
@@ -1,14 +1,30 @@
...
+  const browserVersion = process.env.BROWSER_STACK_BROWSER_VERSION
+  if (notIsNilOrEmpty(browserVersion)) {
```
gethyn1: I probably wouldn't use `notIsNilOrEmpty` here. Javascript string coercion means this is the same as:
```
if (browserVersion) { ... }
```
which I think is a lot less confusing. `notIsNilOrEmpty` is useful when composing functions but I think a simpler approach is better here

src/test/compatibility/custom-json-formatter.js
gethyn1: I guess its a bit confusing that `getBrowserVersion()` is mixed with `isMobileDevice()`. I would maybe split them up where you call `getBrowserVersion()`:
```
...
'version': isMobileDevice() ? '' : getBrowserVersion()
...
```
.. but I'm not sure if that gives you the result you need? If not then I think this is fine.

src/web/routes/application/check/post.js
```diff
@@ -54,6 +54,7 @@ const postCheck = (config) => async (req, res, next) => {
     const err = new Error('Error posting the request')
     err.statusCode = httpStatus.INTERNAL_SERVER_ERROR
     err.error = error.error
+    err.stack += '\nCaused by: ' + error.stack
```
mikeSmith08: This makes sure that the stack trace of the original error is added to the stack of the new error and hence visible in the logs (not completely related to this PR but useful for diagnosing problems contacting the Claimant service).

src/test/compatibility/config/environment.test.js
```diff
@@ -0,0 +1,52 @@
...
+  t.end()
+})
+
+test('getBrowserVersion neither set', (t) => {
+  t.equal(getBrowserVersion(), 'latest', 'should match browser version if both set')
+  resetEnvironment()
+  t.end()
+})
+
+after('teardown', function (t) {
```
gethyn1: Maybe make this an arrow function like the other tests, just for consistency?

src/test/compatibility/custom-json-formatter.js
```diff
@@ -25,7 +24,11 @@ class CustomJsonFormatter extends JsonFormatter {
     pickle,
     scenarioLineToDescriptionMap
   }) {
-    const scenarioData = super.getScenarioData({ featureId: featureId, pickle: pickle, scenarioLineToDescriptionMap: scenarioLineToDescriptionMap })
+    const scenarioData = super.getScenarioData({
+      featureId: featureId,
```
gethyn1: Not required, but these could all use the shorthand syntax:
```
{
  featureId,
  pickle,
  scenarioLineToDescriptionMap
}
```

PR #160, opened 2019-02-19
HTBHF-566 add headers to wiremock claimant service mapping

gethyn1:


src/test/common/wiremock.js
```diff
@@ -1,11 +1,23 @@
 const { postJsonData, performDelete } = require('./request')
 
-const WIREMOCK_MAPPING_URL = process.env.WIREMOCK_URL ? `${process.env.WIREMOCK_URL}/__admin/mappings` : 'http://localhost:8090/__admin/mappings'
+const WIREMOCK_MAPPING_URL = process.env.WIREMOCK_URL
```
gethyn1: I formatted this onto separate lines as it was running off the edge of my screen. Can reset to original state if required.

PR #164, opened 2019-02-20
NO-JIRA Add report timestamp to all Cucumber reports

mikeSmith08:


src/test/acceptance/acceptance-report.js
```diff
@@ -1,12 +1,18 @@
...
 const options = {
   theme: 'bootstrap',
   jsonFile: 'build/reports/acceptance-report.json',
   output: 'build/reports/acceptance-report.html',
   reportSuiteAsScenarios: true,
   brandTitle: 'Acceptance Tests',
-  launchReport: false
+  launchReport: false,
+  metadata: {
+    'Report Run Time': moment(new Date()).format(DATE_FORMAT)
```
YetAnotherMatt: How about 'Tests completed at' instead of 'Report Run Time'?

YetAnotherMatt: Can we add another property that contains the app version (from package.json)?

PR #165, opened 2019-02-20
Feature/standard error logs

gethyn1:
Standardise the format of errors sent to logger by:
- creating an error formatter that merges a custom message with the root error
- format the error message in the global error logging middleware


src/web/routes/application/check/post.js
```diff
@@ -1,7 +1,6 @@
-const httpStatus = require('http-status-codes')
```
gethyn1: I deleted this as we never use `err.statusCode`, but maybe it is required?

src/web/routes/application/common/formatters/errors.js
```diff
@@ -0,0 +1,19 @@
+const mergeErrorMessage = (message, cause) => cause ? `${message}. ${cause.toString()}` : message
+
+const mergeErrorStack = (error, cause) => cause ? `${error.stack}\nCaused by: ${cause.stack}` : error.stack
+
+const mergeErrorProps = ({ cause, message } = {}) => {
+  const error = new Error()
+  Error.captureStackTrace(error, mergeErrorProps)
```
gethyn1: This sets the root of the stack trace as the point at which `mergeErrorProps` was called, e.g. `post.js`

src/web/routes/application/common/formatters/errors.js
```diff
@@ -0,0 +1,19 @@
+const mergeErrorMessage = (message, cause) => cause ? `${message}. ${cause.toString()}` : message
+
+const mergeErrorStack = (error, cause) => cause ? `${error.stack}\nCaused by: ${cause.stack}` : error.stack
+
+const mergeErrorProps = ({ cause, message } = {}) => {
```
YetAnotherMatt: Maybe a name such as 'createError' or 'newError' or similar?

src/web/routes/application/common/formatters/errors.js
```diff
@@ -0,0 +1,20 @@
+const mergeErrorMessage = (message, cause) => cause ? `${message}. ${cause.toString()}` : message
+
+const mergeErrorStack = (error, cause) => cause ? `${error.stack}\nCaused by: ${cause.stack}` : error.stack
+
+const wrapError = ({ cause, message, statusCode }) => {
+  const error = new Error()
```
mikeSmith08: Didn't you say this would cause the error stack to start here and lose the context of where the error was first caught?

src/web/routes/application/common/formatters/errors.test.js
```diff
@@ -0,0 +1,69 @@
...
+})
+
+test('wrapError() returns an error with the correct props when cause is not specified', (t) => {
+  const result = wrapError({ message, statusCode: 500 })
+  const expectedMessage = 'Massive error!!!'
+  const expectedCauseString = '\nCaused by: mock cause error stack'
+
+  t.equal(result instanceof Error, true, 'should be an instance of Error')
+  t.equal(result.message, expectedMessage, 'should have a correctly formatted message property')
+  t.equal(result.stack.includes(expectedCauseString), false, 'should have a correctly formatted stack property')
```
mikeSmith08: The message should be 'should not have a stack included' or similar.

src/web/routes/application/common/formatters/errors.js
gethyn1: This line now deals with that issue:
```
Error.captureStackTrace(error, mergeErrorProps)
```
see https://nodejs.org/api/errors.html#errors_error_capturestacktrace_targetobject_constructoropt

PR #166, opened 2019-02-20
HTBHF-585 Added cookie link in the footer.

dwybourn:


mikeSmith08:
I have no idea why you've done this when I told you yesterday that I was doing this story - we have duplicated effort! I have pretty much exactly the same PR that I was just about to open, I thought you would have noticed that the story was assigned to me already.

src/web/routes/cookies.js
```diff
@@ -0,0 +1,16 @@
+const pageContent = {
+  // TODO setup cookies page to handle translations HTBHF-586
+  cookies: 'Cookies'
```
gethyn1: Should cookies be set as the heading rather than cookies?
```
heading: 'Cookies'
```
.. also, is it a big deal to setup translations for this? otherwise we are merging non translatable content into the master branch, is this OK (maybe its fine)?

src/web/views/templates/page.njk
```diff
@@ -59,6 +60,19 @@
   </div>
 {% endblock %}
 
+{% block footer %}
+  {{ govukFooter({
+    meta: {
+      items: [
+        {
+          href: "/cookies",
+          text: "Cookies"
```
gethyn1: If we are not doing translations in this PR then probably should leave a `TO DO` note to translate here as well

src/web/routes/cookies.js
dwybourn: Had a word with @YetAnotherMatt and decided that no translation would be ok for now.

src/web/views/templates/page.njk
dwybourn: Done.

src/web/routes/cookies.js
dwybourn: Updated to heading

PR #167, opened 2019-02-21
HTBHF-586 Add cookie page content which required moving the getLanguageBase method into a common area.

mikeSmith08:


src/web/routes/cookies.js
```diff
@@ -1,10 +1,13 @@
-const pageContent = {
+const { getLanguageBase } = require('./application/common/language')
+
+const pageContent = ({ language }) => ({
   // TODO setup cookies page to handle translations HTBHF-586
```
YetAnotherMatt: Do we need to add a cy cookie page?

src/web/views/locales/en/cookies.html
```diff
@@ -0,0 +1,87 @@
...
+            <th class="govuk-table__header" scope="col">Name</th>
+            <th class="govuk-table__header" scope="col">Purpose</th>
+            <th class="govuk-table__header" scope="col">Expires</th>
+        </tr>
+    </thead>
+    <tbody class="govuk-table__body">
+        <tr class="govuk-table__row">
+            <td class="govuk-table__cell">htbhf.sid</td>
+            <td class="govuk-table__cell">Used for session management.</td>
+            <td class="govuk-table__cell">When you close your browser.</td>
```
YetAnotherMatt: This will change, but we can update it in HTBHF-483

src/web/routes/cookies.js
mikeSmith08: Yep, that's the next sub-task. There's some work around getting the Cookie link name and translations so thought I'd keep that all together.

src/web/views/locales/en/cookies.html
mikeSmith08: Yes, for now I just copied the copy that I was given by Kate.

PR #171, opened 2019-02-22
HTBHF-HTBHF-611 Adding back links with correct hrefs.

dwybourn:


src/web/routes/application/register-form-routes.js
```diff
@@ -18,13 +18,28 @@ const handleOptionalMiddleware = (operation, fallback = middlewareNoop) => {
   return operation
 }
 
-const createRoute = (csrfProtection, steps, router) => (step) =>
+const getPreviousPage = (steps, step) => {
+  const index = steps.indexOf(step)
+
+  if (index === -1) {
+    throw new Error(`Unable to find ${step} in the list of steps`)
```
YetAnotherMatt: Will this work for the confirmation page?

src/web/routes/application/check/get.js
```diff
@@ -8,7 +8,8 @@ const pageContent = ({ translate }) => ({
   sendApplicationHeader: translate('check.sendApplicationHeader'),
   sendApplicationText: translate('check.sendApplicationText'),
   buttonText: translate('buttons:acceptAndSend'),
-  changeText: translate('check.change')
+  changeText: translate('check.change'),
+  previous: steps[steps.length - 1].path // last step
```
gethyn1: This isn't a safe way to access the `path` property - if steps was empty or undefined, or if the step didn't have a path property for some reason then this would throw an error

src/web/routes/application/register-form-routes.js
dwybourn: This isn't used there.

src/web/routes/application/check/get.js
dwybourn: Updated to use ramda path and check existence of steps before calling .length()

src/web/routes/application/register-form-routes.js
```diff
@@ -18,13 +18,28 @@ const handleOptionalMiddleware = (operation, fallback = middlewareNoop) => {
...
+
+const createRoute = (csrfProtection, steps, router) => (step) => {
+  const previous = getPreviousPage(steps, step)
   router
     .route(step.path)
     .get(
       csrfProtection,
       getSessionDetails,
-      renderView(step.template, step.pageContent, step.next)
+      renderView(step.template, step.pageContent, step.next, previous)
```
gethyn1: I feel like adding `previous` is making the api / surface area for `renderView` too big. I think it would be better to create another middleware step (maybe `configureGet`??) and add the back link to `res.locals`, instead of passing it to the template explicitly in `res.render()`.

PR #176, opened 2019-02-25
HTBHF-486 Navigation acceptance tests

mikeSmith08:
I will not be merging this PR (obviously the tests fail at the moment), I'm just putting this up for an initial review to see if this is along the right lines. Some key things to look at:

- Obviously the `navigate.feature` file, but its the fact that I've done each one as a Scenario Outline. It was much easier to Dev this way, doesn't seem to take that long to run, but I guess its all relative. At the moment all the tests fail so it's difficult to tell how much it adds to the test run.
- How I've done the navigation in `navigation.js` - I was faffing around with nice ways to do this, but in the end reverted to a simple `switch` statement as it was all I could get to work in JavaScript - any other suggestions, please fire away. Also, I was torn as to whether to keep this central in one file or disseminate into the individual page files, but that would require a lot of code from other step files being used in each page files, e.g the card address version would need to know about enter name, enter nino etc etc. Again I'm open to thoughts on this.

Any other thoughts and comments welcome. I will need to change all the ATs that go directly to their page for testing to not short-cut anymore as this will no longer be allowed after this story is competed, so the `navigation.js` steps `I have entered my details up to the (.*) page` will need to be implemented before we do anything further, so I'm keen on comments on this.


src/test/acceptance/features/enter-dob.feature
```diff
@@ -22,7 +22,7 @@ Feature: Date of birth
 
   Scenario Outline: Enter a valid date of birth
     When I enter my date of birth as day: <day>, month: <month> and year: <year>
-    Then the are you pregnant page is shown
+    Then I am shown the are you pregnant page
```
YetAnotherMatt: I like this rewording

src/test/acceptance/features/navigation.feature
```diff
@@ -0,0 +1,52 @@
+Feature: Application process navigation is controlled
+  In order to facilitate the beta HTBHF programme
```
YetAnotherMatt: How about 'In order to help me complete the application correctly'?

src/test/acceptance/features/navigation.feature
```diff
@@ -0,0 +1,52 @@
...
+      | enter name               |
+      | enter national insurance |
+      | enter date of birth      |
+      | are you pregnant         |
+      | card address             |
+
+  Scenario Outline: Navigation after completing application returns to confirm page
+    Given I have completed my application
+    When I navigate to the <page> page
+    Then I am shown a successful confirmation page
```
YetAnotherMatt: We will need some way of allowing applicants to clear their session in this scenario (to support multiple applicants on a shared machine). I'm starting to wonder whether instead we clear their session if they try to go back to other pages?

YetAnotherMatt: I realise that this meets the requirement as stated in the story, mind.

src/test/acceptance/features/navigation.feature
```diff
@@ -0,0 +1,52 @@
...
+      | enter name               |
+      | enter national insurance |
+      | enter date of birth      |
+      | are you pregnant         |
+      | card address             |
+
+  Scenario Outline: Navigation to previous steps in an unsubmitted application flow is allowed
+    Given I have entered my details up to the check details page
+    When I navigate to the <page> page
+    Then I am shown the <page> page
```
YetAnotherMatt: This might be quite a lengthy test. Wish there was a way to loop over steps within a scenario - so that the given was executed once and the when/then once for each example. Short of that I guess we just keep an eye on how long the tests take.

mikeSmith08: Yes, I think I'm more of a mind to keep the tests simple first, and if they get too long then refactor later. On my laptop they seemed to be quite speedy, but I guess it has more hardware than the Travis build servers.

src/test/acceptance/features/navigation.feature
mikeSmith08: Yep, done.

src/test/common/steps/navigation.js
```diff
@@ -0,0 +1,88 @@
...
+  await pages.overview.clickCookieLink()
+})
+
+When(/^I navigate to the (.*) page$/, async function (page) {
+  await navigateToPage(page)
+})
+
+async function navigateToPage (page) {
+  const baseUrl = pages.url
+  switch (page) {
```
gethyn1: The alternative to a switch statement would be to map the `page` to a function - but not sure if its any better:
```
const pageMap = {
  [ENTER_NAME_PAGE]: (baseUrl) => await pages.enterName.openEnterNameUrl(baseUrl),
  [ENTER_NINO_PAGE]: (baseUrl) => pages.enterNino.openEnterNinoUrl(baseUrl)
  ...
}

...

try {
  pageMap[page](baseUrl)
} catch (error) {
  throw new Error(...)
}
```

src/test/common/steps/navigation.js
```diff
@@ -0,0 +1,66 @@
...
+      break
+    case ENTER_NINO_PAGE:
+      await enterNameAndSubmit(FIRST_NAME, LAST_NAME)
+      break
+    case ENTER_DOB_PAGE:
+      await enterNameAndSubmit(FIRST_NAME, LAST_NAME)
+      await enterNinoAndSubmit(VALID_NINO)
+      break
+    case ARE_YOU_PREGNANT_PAGE:
+      await enterNameAndSubmit(FIRST_NAME, LAST_NAME)
```
dwybourn: To avoid duplication you can reverse this switch statement and remove the break statements. Then each switch just needs the one statement.

src/test/common/page/page.js
```diff
@@ -34,7 +43,7 @@ class Page {
     }
   }
 
-  async openDirect (url, lang) {
+  getPath () {
     throw new Error('openDirect needs to be implemented by all Page objects')
```
dwybourn: Error message is wrong.

PR #177, opened 2019-02-25
HTBHF-284 Added acceptance tests and back link to cookies page.

dwybourn:


src/test/acceptance/features/cookies.feature
```diff
@@ -0,0 +1,15 @@
+Feature: Cookies details page
+  In order to apply for the HTBHF programme
+  As a potential claimant
+  I want to view the cookies used by the site
+
+  Scenario: The cookie back link is not shown if we navigate directly to it
```
mikeSmith08: I think this would be better as `The back link on the Cookies page is not shown if we navigate to it directly`.

src/test/acceptance/features/cookies.feature
```diff
@@ -0,0 +1,15 @@
...
+    When I go directly to the cookies page
```
mikeSmith08: I'm not sure the word `directly` adds anything here.

src/test/acceptance/features/cookies.feature
```diff
@@ -0,0 +1,15 @@
...
+    Then the cookies page is shown
+    And no back link is shown
+
+  Scenario: The cookie back link is shown and links to the last page in the application if we navigate to it via a link in the application
```
mikeSmith08: How about: `The back link on the Cookie page is shown when the page is accessed via an application page`

src/test/common/page/page.js
```diff
@@ -133,6 +134,15 @@ class Page {
     const cookieLink = await this.findByCSS(COOKIES_LINK_CSS)
     await cookieLink.click()
   }
+
+  async getBackLink () {
+    try {
+      return this.findByClassName(BACK_LINK_CLASSNAME)
+    } catch (error) {
+      // throw error back up to test, needed for absence of link test.
```
mikeSmith08: If you're intentionally doing this, what's the point in having the try/catch block at all if you just throw that error?

src/test/common/steps/cookie-steps.js
```diff
@@ -1,10 +1,18 @@
...
 const pages = require('./pages')
 
 const GOV_UK_TABE_CLASSNAME = 'govuk-table'
 
+When(/^I go directly to the cookies page$/, async function () {
+  await pages.cookies.open(pages.url)
+})
+
+When(/^I click on the cookies link$/, async function () {
+  await pages.enterName.clickCookieLink()
```
mikeSmith08: You can access this directly from `genericPage`, e.g. `pages.genericPage.clickCookieLink()`.

src/test/common/steps/cookie-steps.js
```diff
@@ -13,6 +21,23 @@ Then(/^all page content is present on the cookies page$/, async function () {
   await checkAllCookiePageContentIsPresentAndCorrect()
 })
 
+Then(/^no back link is shown$/, async function () {
+  try {
+    console.log('Timeout error for getting back link expected')
+    await pages.cookies.getBackLink()
+    assert.fail('Backlink found on page but was not expected')
+  } catch (error) {
+    // no nothing, we expect there not to be a back link.
```
mikeSmith08: **Do** nothing? Also being picky, but `we expect there to be no back link` is better.

src/test/common/steps/cookie-steps.js
```diff
@@ -13,6 +21,23 @@ Then(/^all page content is present on the cookies page$/, async function () {
...
+    return true
+  }
+})
+
+Then(/^the back link on the cookies page links to the enter name page$/, async function () {
+  const backLink = await pages.cookies.getBackLink()
+  const href = await backLink.getAttribute('href')
+  expect(href).to.equal(`${pages.url}/enter-name`)
```
mikeSmith08: I think this URL should be a constant in the `EnterName` page object as it is already defined and used in there, so safer to match via a constant.

mikeSmith08:
This is going to clash majorly with my acceptance test branch `feature/`navigationAts`, we should coordinate before either of these two branches are merged.

src/test/acceptance/features/cookies.feature
dwybourn: I would in this case as how you got to the page is the key part of the feature.

src/test/common/page/page.js
dwybourn: This has been removed.

src/test/common/steps/cookie-steps.js
dwybourn: Updated

src/test/common/steps/cookie-steps.js
dwybourn: Updated text.

src/test/common/steps/cookie-steps.js
dwybourn: Done.

src/test/acceptance/features/cookies.feature
dwybourn: Done.

dwybourn: Done.

src/test/acceptance/features/cookies.feature
dwybourn: Done

src/test/common/page/page.js
dwybourn: This has been removed.

src/test/common/steps/cookie-steps.js
dwybourn: Changed.

src/test/common/steps/cookie-steps.js
dwybourn: Done.

src/test/common/steps/cookie-steps.js
```diff
@@ -13,12 +21,32 @@ Then(/^all page content is present on the cookies page$/, async function () {
   await checkAllCookiePageContentIsPresentAndCorrect()
 })
 
+Then(/^no back link is shown$/, async function () {
+  try {
+    console.log('Timeout error for getting back link expected')
+    await pages.cookies.getBackLink()
```
YetAnotherMatt: Is there any way we can do this without a timeout?

src/test/acceptance/features/cookies.feature
```diff
@@ -8,3 +8,14 @@ Feature: Cookies page
     When I click the Cookies link
     Then the cookies page is shown
     And all page content is present on the cookies page
+
+  Scenario: The back link on the Cookies page is not shown if we navigate to it directly
+    When I go directly to the cookies page
```
mikeSmith08: You should now be using the same language as our new tests, so this should be `I navigate to the cookies page`, there is already a step you can use.

src/test/common/steps/cookie-steps.js
```diff
@@ -1,9 +1,17 @@
...
-const { expect } = require('chai')
+const { When, Then } = require('cucumber')
+const { assert, expect } = require('chai')
 
 const pages = require('./pages')
 
-const GOV_UK_TABE_CLASSNAME = 'govuk-table'
+const GOV_UK_TABLE_CLASSNAME = 'govuk-table'
+
+When(/^I go directly to the cookies page$/, async function () {
```
mikeSmith08: You don't need this, use `I navigate to the cookies page` which is in `navigation.js`.

src/web/routes/application/middleware/set-previous-page.js
```diff
@@ -12,13 +12,13 @@ const getPreviousPage = (steps, step) => {
   }
 }
 
-const configureGet = (steps, step) => (req, res, next) => {
+const setPreviousPage = (steps, step) => (req, res, next) => {
   res.locals.previous = getPreviousPage(steps, step)
```
gethyn1: Is it worth just calling this inside `configureGet` rather than creating a new piece of middleware for setting this single value?

dwybourn: Moved setPreviousPage into configureGet/Post

src/test/common/steps/cookie-steps.js
dwybourn: Removed.

src/test/acceptance/features/cookies.feature
dwybourn: Updated.

PR #179, opened 2019-02-26
NO-JIRA Doubled the amount of memory available to instances

YetAnotherMatt:


YetAnotherMatt:
The amount of memory also governs the quota of vCPU available to the instance - so this will double the cpu quota and hopefully reduce the performance fluctuations

PR #183, opened 2019-02-28
Feature/holding page

dwybourn:


src/web/routes/register-routes.js
```diff
@@ -21,11 +24,17 @@ const registerRoutes = (config, app) => {
...
   app.use(setCommonTemplateValues)
 
-  registerFormRoutes(csrfProtection, steps, app)
-  registerStartRoute(app)
-  registerCheckRoutes(csrfProtection, config, app)
-  registerConfirmRoute(app)
-  registerCookiesRoute(app)
+  const maintenanceMode = getVCAPServicesVariable('MAINTENANCE_MODE', false)
+
+  if (toBoolean(maintenanceMode)) {
```
YetAnotherMatt: Can this be unit tested?

src/web/routes/register-routes.js
```diff
@@ -21,11 +24,17 @@ const registerRoutes = (config, app) => {
 
   app.use(setCommonTemplateValues)
 
-  registerFormRoutes(csrfProtection, steps, app)
-  registerStartRoute(app)
-  registerCheckRoutes(csrfProtection, config, app)
-  registerConfirmRoute(app)
-  registerCookiesRoute(app)
+  const maintenanceMode = getVCAPServicesVariable('MAINTENANCE_MODE', false)
```
gethyn1: Is it possible to set `maintenanceMode` on the `config` object that is passed down from app initialisation? `config` is already available inside `registerRoutes` .. it's the approach that we've taken with the rest of the app.

dwybourn: Moved into config.

PR #188, opened 2019-03-04
Feature/redirect when invalid path is requested

gethyn1:
- Create middleware to validate user is requesting an allowed path in the apply journey
- Redirects to next allowed path in the apply journey if requested path is not allowed

### DIFF
--- a/src/test/acceptance/features/navigation.feature
+++ b/src/test/acceptance/features/navigation.feature
@@ -15,6 +15,7 @@ Feature: Application process navigation is controlled
       | enter date of birth      |
       | are you pregnant         |
       | card address             |
+      | check details            |
       | confirmation             |
 
   @ignore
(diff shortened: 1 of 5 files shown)


src/test/acceptance/features/navigation.feature
```diff
@@ -32,16 +33,26 @@ Feature: Application process navigation is controlled
...
+      | enter national insurance | check details   |
+      | enter date of birth      | check details   |
+      | are you pregnant         | check details   |
+      | card address             | check details   |
+      | enter name               | confirmation    |
+      | enter national insurance | confirmation    |
+      | enter date of birth      | confirmation    |
+      | are you pregnant         | confirmation    |
+      | card address             | confirmation    |
+
```
YetAnotherMatt: This looks like it will add quite a bit to our test time. Gethyn says the plan may be to test with this many initially, then cut the list back a bit?

src/web/routes/application/middleware/handle-path-request.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  if (!isPathAllowed(pathsInSequence, req.session.nextAllowedStep, req.path)) {
+    return res.redirect(req.session.nextAllowedStep)
+  }
+
+  next()
+}
+
+const handlePathRequest = (steps) => middleware(getPathsInSequence(steps))
```
gethyn1: I have discussed with @YetAnotherMatt maybe changing the name of this to `handleRequestForPath` to stop any confusion that this is related to HTTP `request`. @mikeSmith08 how do you feel about that?

src/test/acceptance/features/navigation.feature
mikeSmith08: Yes, exactly. For now, this will be quite useful, but I think it would be fine to cut this down to about 3 examples and it would give us a good enough level of coverage.

src/web/routes/application/middleware/handle-path-request.js
mikeSmith08: Yes, I'm happy with that.

PR #190, opened 2019-03-04
Feature/path check final routes

gethyn1:
- Redirect requests to invalid paths for /check and /confirm routes
- Move `createRequestBody()` to separate file
- Add happy path unit tests for check details POST request

### DIFF
--- a/src/test/acceptance/features/navigation.feature
+++ b/src/test/acceptance/features/navigation.feature
@@ -3,7 +3,6 @@ Feature: Application process navigation is controlled
   As a potential claimant
   I want to make sure that only the correct application flow can be used
 
-  @ignore
   Scenario Outline: Enter name page is always the first page
     Given I am starting a new application
     When I navigate to the <page> page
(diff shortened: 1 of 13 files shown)


src/test/acceptance/features/navigation.feature
```diff
@@ -31,7 +30,6 @@ Feature: Application process navigation is controlled
       | are you pregnant         |
       | card address             |
 
-  @ignore
   Scenario Outline: Navigation is not allowed past the current page in the flow
```
YetAnotherMatt: Presumably we'll be removing most of the examples in this scenario?

src/web/routes/application/check/create-request-body.js
```diff
@@ -0,0 +1,33 @@
+const { toDateString } = require('../common/formatters')
```
gethyn1: There is no new code here - just moved into a new file to reduce the size of `post.test.js`

src/test/acceptance/features/navigation.feature
mikeSmith08: Yep, I've left one per navigation page now, that should cover us sufficiently.

PR #192, opened 2019-03-05
HTBHF-624 Added acceptance tests for generic error page.

dwybourn:


src/test/acceptance/features/error-page.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Sever Error page
```
gethyn1: typo - I guess this should be 'Server'

src/test/acceptance/features/error-page.feature
```diff
@@ -0,0 +1,9 @@
...
+  In order to facilitate the beta HTBHF programme
```
gethyn1: Should we be using `beta`? I'm assuming we will still be using this code once the beta is complete?

src/test/acceptance/features/error-page.feature
dwybourn: Fixed typo.

src/test/acceptance/features/error-page.feature
dwybourn: We have this in most of the feature files. I would leave it in for now as the program is beta.

PR #203, opened 2019-03-26
HTBHF-503 Created a privacy notice page

YetAnotherMatt:


src/test/acceptance/features/privacy-notice.feature
```diff
@@ -0,0 +1,21 @@
+Feature: Privacy notice page
```
gethyn1: Should this be *privacy policy*? Or is that the agreed language? *privacy notice* sounds like a popup or banner (subjective I guess).

src/test/acceptance/features/privacy-notice.feature
```diff
@@ -0,0 +1,21 @@
...
+  In order to facilitate the beta HTBHF programme
+  As a potential claimant
+  I want to see the details of how my details are used and retained
+
+  Scenario: The Privacy notice page is accessible
+    Given I navigate to the HTBHF overview page
```
mikeSmith08: Might it be prudent for this to not be the Overview page and rather the enter name page as the Overview page will disappear at some point?

src/test/acceptance/features/privacy-notice.feature
YetAnotherMatt: That's what it is called in the prototype - i.e. in the content from Kate

src/test/acceptance/features/privacy-notice.feature
YetAnotherMatt: Fair point. There is then no difference between this and the last scenario - shall I remove this one? The same would apply for the cookies feature.

mikeSmith08: Yes, I believe so.

src/test/common/steps/privacy-notice-steps.js
```diff
@@ -0,0 +1,38 @@
...
+  await checkPrivacyNoticePageContentIsPresentAndCorrect()
+})
+
+Then(/^the back link on the privacy-notice page links to the enter name page$/, async function () {
+  const backLink = await pages.privacyNotice.getBackLink()
+  const href = await backLink.getAttribute('href')
+
+  // using startsWith() instead of equals() so query parameters do not fail the test
+  const enterNameUrl = `${pages.url}${pages.enterName.getPath()}`
+  expect(href.startsWith(enterNameUrl)).to.equal(true)
```
mikeSmith08: You can use Chai's `startWith`, so something like: `expect('abcdef').to.startsWith('abc')` or `'abcdef'.should.startWith('abc')` if the grammar of the first doesn't seem quite right.

YetAnotherMatt: Gethyn thinks it requires bringing in a new package - not worthwhile for this one assertion

mikeSmith08: That's fine.

PR #205, opened 2019-04-02
HTBHF-640 Handle 404 responses from claimant service.

dwybourn:
Note, additional acceptance tests and content will be created in a future PR. This PR is up so that the service doesn't break when the api changes for the claimant service goes in.

### DIFF
--- a/src/web/routes/application/check/post.js
+++ b/src/web/routes/application/check/post.js
@@ -22,19 +22,21 @@ const postCheck = (steps, config) => async (req, res, next) => {
         claimant: createRequestBody(req.session.claim)
       }
     })
-
-    logger.info('Sent claim', { req })
-
-    stateMachine.setState(states.COMPLETED, req)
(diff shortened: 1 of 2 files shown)


src/web/routes/application/check/post.js
```diff
@@ -22,19 +22,21 @@ const postCheck = (steps, config) => async (req, res, next) => {
...
-    stateMachine.setState(states.COMPLETED, req)
-    req.session.nextAllowedStep = stateMachine.dispatch(actions.GET_NEXT_PATH, req, steps, req.path)
-    return res.redirect('confirm')
   } catch (error) {
-    next(wrapError({
-      cause: error,
-      message: 'Error posting to claimant service',
-      statusCode: httpStatus.INTERNAL_SERVER_ERROR
-    }))
+    if (error.statusCode !== 404) {
```
gethyn1: We might be able to update the `request` options argument using `resolveWithFullResponse` or `simple` to handle this in the try block. Checking the response status in the catch block doesn't feel quite right to me.

https://github.com/request/request-promise#fulfilled-promises-and-the-resolvewithfullresponse-option

PR #206, opened 2019-04-02
Handle not found response from claimant service outside of error handler

gethyn1:
Add a response transformer to treat 404 responses from the claimant service as successful

### DIFF
--- a/src/web/routes/application/check/post.js
+++ b/src/web/routes/application/check/post.js
@@ -1,5 +1,6 @@
 const httpStatus = require('http-status-codes')
 const request = require('request-promise')
+const { path } = require('ramda')
 
 const { wrapError } = require('../common/formatters')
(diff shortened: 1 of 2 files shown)


src/web/routes/application/check/post.js
```diff
@@ -9,36 +10,52 @@ const { createRequestBody } = require('./create-request-body')
...
+    },
+    body: {
+      claimant: createRequestBody(req.session.claim)
+    },
+    simple: false,
+    transform: transformResponse
+  })
+    .then(
+      () => {
+        logger.info('Sent claim', { req })
```
YetAnotherMatt: Can we move this log line to before we send the request? (`logger.info('Sending claim', { req })`) As we log a change of state in the next line, we'll then get log events before and after making the rest call.

PR #210, opened 2019-04-03
HTBHF-924 Add test:integration target for e2e integration test

gethyn1:
Add `test:integration` target to run the compatibility tests against a local target in headless Chrome. This can be used as an e2e integration test in the CD pipeline.


YetAnotherMatt:
test:integration will need to run against staging, so not point to a local target. It will need to run in the same way that smoke tests do - being provided a url as an environment variable.

mikeSmith08:
@YetAnotherMatt We've now changes the config for the Integration tests to use `APP_BASE_URL`. As for the second part, I don't think we need to do this because it runs the same test as the compatibility tests and we already run `compatibility:test:local` as a part of the build, so this would be just a duplicate.

PR #214, opened 2019-04-10
Update vulnerable npm packages

gethyn1:
- Updated `i18next-node-fs-backend` to fix `js-yaml` vulnerability
- Updated Nunjucks to ensure using latest version of node-gyp
- `node-gyp` imports vulnerable version of `tar`; listed `tar` as project dependency to deal with this. Following issue https://github.com/nodejs/node-gyp/issues/1714 so hopefully can remove this dependency in the future

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2708,6 +2708,14 @@
         "universalify": "^0.1.0"
       }
     },
+    "fs-minipass": {
+      "version": "1.2.5",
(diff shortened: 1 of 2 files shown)


stof:
this change does not prevent node-gyp from installing an older version of tar. You simply have multiple version of tar available in your node_modules folder now.

gethyn1:
@stof Thanks - yes we realised that this morning. I was mistakenly under the impression that there was only ever one version of a package installed.

PR #218, opened 2019-04-17
HTBHF-978 assert on total voucher value for four weeks

gethyn1:
The content of the confirm page is liable to change as we iterate the apply journey. When this content changes the function `assertVoucherEntitlementValuesForAttr()` will allow quick updates to the acceptance tests by wrapping voucher entitlement values in `data` attributes e.g:
```
<span data-entitlement="total-voucher-value-four-weeks">{{ totalVoucherValueForFourWeeks }}</span>
```
The above HTML was introduced in a previous commit in `src/web/views/locales/en/confirm.html`. This PR asserts that the total voucher value for four weeks is correct.

### DIFF
--- a/src/test/common/steps/confirm-details.js
+++ b/src/test/common/steps/confirm-details.js
@@ -23,6 +23,7 @@ Then(/^I am shown a successful confirmation page$/, async function () {
 Then(/^my entitlement is 12.40 per week$/, async function () {
   const totalVoucherValue = await pages.confirm.getPanelBodyText()
   expect(totalVoucherValue.toString().trim()).to.be.equal('You are entitled to\n£12.40 per week', 'expected total voucher value to be correct')
+  await assertVoucherEntitlementValuesForAttr('total-voucher-value-four-weeks', '49.60')
 })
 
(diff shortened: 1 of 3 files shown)


src/test/common/steps/confirm-details.js
```diff
@@ -31,3 +32,14 @@ async function checkAllPageContentIsPresentAndCorrect () {
   const panelTitle = await pages.confirm.getPanelTitleText()
   expect(panelTitle.toString().trim()).to.be.equal('Application complete', 'expected confirmation header to be correct')
 }
+
+async function assertVoucherEntitlementValuesForAttr (attr, expectedValue) {
+  try {
+    const message = `expected all references to ${attr.replace('-', ' ')} to be correct`
+    const refs = await pages.confirm.getVoucherEntitlementRefsForDataAttr(attr)
+    const incorrectAmounts = refs.filter(value => value !== expectedValue)
+    expect(incorrectAmounts.length).to.be.equal(0, message)
```
mikeSmith08: Is there a `expected(incorrectAmounts).to.be.empty(message)` that we can use here instead?

PR #230, opened 2019-06-03
Feature/updated claim

dwybourn:


src/web/routes/application/check/post.test.js
```diff
@@ -81,14 +81,20 @@ test(`successful post sets next allowed step to ${CONFIRM_URL} and eligibility s
 
   post.returns(Promise.resolve({
     body: {
-      eligibilityStatus: ELIGIBLE
+      eligibilityStatus: ELIGIBLE,
```
YetAnotherMatt: shouldn't we also include claimStatus: ACTIVE here?

dwybourn: The claimStatus isn't added to the session so isn't needed here.

src/test/common/steps/complete-flow.js
```diff
@@ -49,3 +53,11 @@ async function submitApplicationWithStatus (status) {
   await setupWiremockMappingsWithStatus(status)
   await pages.genericPage.submitForm()
 }
+
+async function submitApplicationForUpdatedClaim () {
```
mikeSmith08: This function is initially confusing until you realise that the flow in the application is identical for the update process as to the new applicant process, all we're doing is simulating a different response from Wiremock to mimic that the claim was updated. It feels as though we should do better, but we simply can't. Oh well.

src/test/common/page/confirm-updated.js
```diff
@@ -0,0 +1,15 @@
+'use strict'
+
+const Confirm = require('./confirm')
+const CONFIRM_UPDATED_PAGE_TITLE = 'GOV.UK - Application updated'
+
+/**
+ * Page object for the confirmation page after submitting the claim.
```
mikeSmith08: Might be worth mention that this is for the update flow, but its the same page as the normal confirm flow.

src/test/common/steps/confirm-details.js
```diff
@@ -33,6 +42,13 @@ async function checkAllPageContentIsPresentAndCorrect () {
   expect(panelTitle.toString().trim()).to.be.equal('Application complete', 'expected confirmation header to be correct')
 }
 
+async function checkExistingClaimUpdatedTextIsPresent () {
```
mikeSmith08: I'm guessing we have no different content at all yet? If so, might be worthwhile adding a TODO with a Jira number in it for reference.

src/test/common/steps/pages.js
```diff
@@ -61,6 +63,7 @@ class Pages {
     this.cardAddress = new CardAddress(this.driver)
     this.check = new Check(this.driver)
     this.confirm = new Confirm(this.driver)
+    this.confirmUpdated = new ConfirmUpdated(this.driver)
```
mikeSmith08: You might want to add it to the `allPages` map too in case it diverges from the Confirm page at any point.

src/test/common/wiremock.js
```diff
@@ -52,10 +55,40 @@ const createSuccessfulClaimsMapping = (eligibilityStatus) => {
...
+      }
+    }
+  })
+}
+
 const ERROR_CLAIMS_MAPPING = JSON.stringify({
   'request': {
     'method': 'POST',
-    'url': '/v1/claims',
+    'url': '/v2/claims',
```
mikeSmith08: Would it be worthwhile having this as a constant seeing as you had up date it 3 times?

mikeSmith08: I see there is already one, so let's use it.

src/test/common/steps/pages.js
dwybourn: Done.

src/test/common/steps/confirm-details.js
dwybourn: Done.

src/test/common/page/confirm-updated.js
dwybourn: Done.

src/test/common/wiremock.js
dwybourn: Added in the test constants file and using that.

PR #232, opened 2019-06-10
Reorder UI sequence

gethyn1:
- Reorder sequence of questions in UI to match prototype
- Update browser tests to match reordered UI sequence
- Refactor internationalisation feature to avoid renaming of scenarios as new steps are added to the start of the application
- Create a single `startApplication()` function to avoid multiple updates as new pages are added to the start of the application

### DIFF
--- a/src/test/acceptance/features/internationalisation.feature
+++ b/src/test/acceptance/features/internationalisation.feature
@@ -4,10 +4,10 @@ Feature: Translations
   I want to view the site in my preferred language
 
   Scenario:
-    When I have entered my details up to the enter national insurance page with Welsh language selected
-    Then the enter national insurance page is in Welsh
+    When I have completed the first step of the application with Welsh language selected
+    Then the next page is displayed in Welsh
 
   Scenario:
-    Given I have entered my details up to the enter national insurance page with Welsh language selected
-    When I enter a valid national insurance number
-    Then I am on the next page which is displayed in Welsh
+    Given I have completed the first step of the application with Welsh language selected
+    When I succesfully complete the next step
+    Then the next page continues to be displayed in Welsh
--- a/src/test/acceptance/features/navigation.feature
+++ b/src/test/acceptance/features/navigation.feature
@@ -3,10 +3,10 @@ Feature: Application process navigation is controlled
   As a potential claimant
   I want to make sure that only the correct application flow can be used
 
-  Scenario Outline: Enter name page is always the first page
+  Scenario Outline: Enter date of birth page is always the first page
     Given I am starting a new application
     When I navigate to the <page> page
-    Then I am shown the enter name page
+    Then I am shown the enter date of birth page
     Examples:
       | page                     |
       | enter name               |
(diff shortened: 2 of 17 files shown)


src/test/acceptance/features/internationalisation.feature
```diff
@@ -4,10 +4,10 @@ Feature: Translations
   I want to view the site in my preferred language
 
   Scenario:
-    When I have entered my details up to the enter national insurance page with Welsh language selected
-    Then the enter national insurance page is in Welsh
+    When I have completed the first step of the application with Welsh language selected
```
gethyn1: I have renamed the scenarios in this file to be more generic to avoid constantly renaming as new steps are added to the start of the application. In my opinion the names of the pages are not important in this feature, just that a sequence of pages are loaded in the correct language.

src/test/acceptance/features/navigation.feature
```diff
@@ -30,10 +30,10 @@ Feature: Application process navigation is controlled
...
       | page                     |
-      | enter name               |
-      | enter national insurance |
       | enter date of birth      |
       | are you pregnant         |
+      | enter name               |
+      | enter national insurance |
       | card address             |
 
   Scenario Outline: Navigation is not allowed past the current page in the flow
```
gethyn1: I feel like the examples (needs expanding to view code) in this feature should not work as they have not been updated, but nothing is failing. Anyone got any thoughts on this?

src/test/common/steps/pages.js
```diff
@@ -79,6 +79,10 @@ class Pages {
   async openPageDirect (pageName) {
     await this.pageMap[pageName](this.url)
   }
+
+  async startApplication () {
+    this.enterDOB.open(this.url)
```
gethyn1: It seemed sensible to create a start application function to avoid multiple updates every time the first page is changed. This seemed like a sensible place as it already has access to the URL, but maybe it should be a function in common steps?

src/test/acceptance/features/navigation.feature
dwybourn: This feature doesn't care about the order of the pages, just the state of the application, i.e IN_REVIEW or COMPLETE.

src/test/acceptance/features/navigation.feature
```diff
@@ -3,16 +3,16 @@ Feature: Application process navigation is controlled
...
     When I navigate to the <page> page
-    Then I am shown the enter name page
+    Then I am shown the enter date of birth page
     Examples:
       | page                     |
-      | enter name               |
-      | enter national insurance |
       | enter date of birth      |
       | are you pregnant         |
+      | enter name               |
```
mikeSmith08: I don't think you needed to make this change as each example runs a new test and the order between tests doesn't make any difference, same with the change below too.

src/test/acceptance/features/overview.feature
```diff
@@ -6,12 +6,12 @@ Feature: Temporary overview page
   Scenario: The overview page is available and starts the application process
     Given I navigate to the HTBHF overview page
     When I select to start the process
-    Then I am shown the enter name page
+    Then I am shown the enter date of birth page
```
mikeSmith08: As per the comments about keeping this more generic, you could call this `Then I am shown the first page of the application process`. I also have no objection to keeping it like this either.

PR #238, opened 2019-06-13
HTBHF-1557 add validation for do you live in Scotland

gethyn1:
- Add validation to "Do you live in Scotland?"
- Shared code from "Are you pregnant?" has been refactored into common functions

### DIFF
--- a/src/test/acceptance/features/check-details.feature
+++ b/src/test/acceptance/features/check-details.feature
@@ -25,7 +25,7 @@ Feature: Check details
   Scenario: I can change my answer to a question and am returned to the check details page with the correct information
     Given I am on the check details page having entered valid details for a pregnant woman
     When I choose to change my answer to are you pregnant
-    And I select the No option on the are you pregnant page
+    And I say No to the are you pregnant question
     And I click continue
     Then I am shown the check details page
     And the check details page contains all data entered for a woman who is not pregnant
--- a/src/test/acceptance/features/check-details.feature
+++ b/src/test/acceptance/features/check-details.feature
@@ -25,7 +25,7 @@ Feature: Check details
   Scenario: I can change my answer to a question and am returned to the check details page with the correct information
     Given I am on the check details page having entered valid details for a pregnant woman
     When I choose to change my answer to are you pregnant
-    And I select the No option
+    And I select the No option on the are you pregnant page
     And I click continue
     Then I am shown the check details page
     And the check details page contains all data entered for a woman who is not pregnant
--- a/src/test/acceptance/features/do-you-live-in-scotland.feature
+++ b/src/test/acceptance/features/do-you-live-in-scotland.feature
@@ -1,0 +1,20 @@
+Feature: Do you live in Scotland?
+  In order to apply for the HTBHF programme
+  As a potential claimant
+  I want to select whether or not I live in Scotland
+
+  Background:
+    Given I have entered my details up to the do you live in scotland page
+
+  Scenario: Yes and No are displayed
+    Then Yes and No options are displayed
+
+  Scenario: Do not select an option
+    When I do not select an option
+    And I click continue
+    Then I am informed that I need to select an option for do you live in scotland
+
+  Scenario: Select the No option
+    When I select the No option
+    And I click continue
+    Then I am shown the enter date of birth page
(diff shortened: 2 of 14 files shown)


src/test/acceptance/features/do-you-live-in-scotland.feature
```diff
@@ -0,0 +1,20 @@
+Feature: Do you live in Scotland?
+  In order to apply for the HTBHF programme
+  As a potential claimant
+  I want to select whether or not I live in Scotland
+
+  Background:
+    Given I have entered my details up to the do you live in scotland page
+
+  Scenario: Yes and No are displayed
+    Then Yes and No options are displayed
```
dwybourn: I'm surprised this works as it's only the are you pregnant steps that defined this. I would update this to include the page name and add that to the do you live in scotland steps.

src/web/routes/application/do-you-live-in-scotland/validate.js
```diff
@@ -0,0 +1,11 @@
+const { check } = require('express-validator/check')
+const { translateValidationMessage } = require('../common/translate-validation-message')
+const { YES, NO } = require('../common/constants')
+
+const validate = [
+  check('doYouLiveInScotland').isIn([YES, NO]).withMessage(translateValidationMessage('validation:selectYesOrNo'))
```
dwybourn: Could this be moved into somewhere common? I imagine it's the same of are you pregnant?

src/test/acceptance/features/do-you-live-in-scotland.feature
mikeSmith08: It works fine because it simply calls `findElements` on `WebDriver` so the fact that it's on the Are You  Pregnant page doesn't matter. Either way, I'll move it out into common steps.

mikeSmith08: There were a few other steps that needed to be sorted too, but this has now been done.

src/web/routes/application/do-you-live-in-scotland/validate.js
gethyn1: This is a fair point but I would say that it is unnecessary abstraction, adding more problems (in terms of indirection during debugging) than it solves - if most validation have inline messages but this abstraction breaks that rule, it makes mentally parsing the code that bit more difficult.

It would require a function that takes in the field which would in turn require adding to the validation array:
```
const checkIsYesOrNo = (fieldName) => ...

const validate = [
  ...checkIsYesOrNo('doYouLiveInScotland'),
  check('SomethingElse').withMessage('message is right here'),
  check('theNextThing').withMessage('message is right here')
]
```
I don't think it's a massive overhead to have a small amount of duplication in this code.

gethyn1: Having said that, its not a difficult change so happy to make the update if others don't agree with me.

src/test/acceptance/features/check-details.feature
```diff
@@ -25,7 +25,7 @@ Feature: Check details
   Scenario: I can change my answer to a question and am returned to the check details page with the correct information
     Given I am on the check details page having entered valid details for a pregnant woman
     When I choose to change my answer to are you pregnant
-    And I select the No option
+    And I select the No option on the are you pregnant page
```
YetAnotherMatt: This might be better worded as 'And I say No to the are you pregnant question'. The scenario already makes it clear what page we're on

src/test/acceptance/features/do-you-live-in-scotland.feature
```diff
@@ -0,0 +1,20 @@
...
+  Scenario: Yes and No are displayed
+    Then Yes and No options are displayed on the do you live in scotland page
+
+  Scenario: Do not select an option
+    When I do not select an option
+    And I click continue
+    Then I am informed that I need to select an option for do you live in scotland
+
+  Scenario: Select the No option
+    When I select the No option on the do you live in scotland page
```
YetAnotherMatt: Similarly this could be `When I say No to the do you live in Scotland question`. 
Also, Scotland should be capitalised - all instances on this page need to be capitalised.

src/web/routes/application/do-you-live-in-scotland/validate.js
gethyn1: ... on second thoughts I have addressed this with a refactor!!

src/test/acceptance/features/check-details.feature
mikeSmith08: Done.

src/test/acceptance/features/do-you-live-in-scotland.feature
mikeSmith08: Done.

PR #240, opened 2019-06-14
Allow conditional next paths for step

gethyn1:
- Create a function that accepts the request object and returns a string in order to determine a path for the next step. This allows for conditional rendering of the next step in the sequence.
- Update the state machine to implement this function when determining the next step for the `IN_PROGRESS` state.
- Update the steps configuration objects to use a function for the `next` property.

### DIFF
--- a/src/web/routes/application/common/state-machine/get-next-for-step.js
+++ b/src/web/routes/application/common/state-machine/get-next-for-step.js
@@ -1,0 +1,25 @@
+const { path } = require('ramda')
+
+const getNextForStep = (req, step) => {
+  const next = path(['next'], step)
+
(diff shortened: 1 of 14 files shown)


src/web/routes/application/common/state-machine/get-next-for-step.js
```diff
@@ -0,0 +1,25 @@
+const { path } = require('ramda')
+
+const getNextForStep = (req, step) => {
+  const next = path(['next'], step)
+
+  if (typeof next !== 'function') {
```
dwybourn: Would `next instanceof Function` read better?

PR #246, opened 2019-06-19
Feature/i live in scotland ats

mikeSmith08:
- Check whether the request is to a navigable step, if not go to next step.
- Add acceptance tests for I Live In Scotland page and navigation.

### DIFF
--- a/src/test/common/page/i-live-in-scotland.js
+++ b/src/test/common/page/i-live-in-scotland.js
@@ -8,7 +8,7 @@ const PAGE_TITLES = {
 }
 
 /**
- * Page object for ILiveInScotland page where the claimant is told they cannot apply if they live in Scotland.
+ * Page object for I Live in Scotland page where the claimant is told they cannot apply if they live in Scotland.
  */
 class ILiveInScotland extends Page {
   async waitForPageLoad (lang = 'en') {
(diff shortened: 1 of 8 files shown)


src/test/common/page/i-live-in-scotland.js
```diff
@@ -0,0 +1,38 @@
...
+
+const Page = require('./page')
+
+const PAGE_TITLES = {
+  en: 'GOV.UK - You cannot apply if you live in Scotland',
+  cy: 'GOV.UK - Sed do eiusmod tempor incididunt ut labore et dolore'
+}
+
+/**
+ * Page object for ILiveInScotland page where the claimant is told they cannot apply if they live in Scotland.
```
gethyn1: `Page object for ILiveInScotland` -> `Page object for I live in Scotland`

src/web/routes/application/middleware/handle-path-request.js
```diff
@@ -26,10 +28,15 @@ const middleware = (config, pathsInSequence) => (req, res, next) => {
     return res.redirect(nextAllowedPath)
   }
 
+  // If step is not navigable then return to the next allowed path (which will be the originating page)
```
gethyn1: I don't think this will always be the originating page - the referring path and `nextAllowedPath` could be different

PR #247, opened 2019-06-19
HTBHF-1377 include app version when sending a claim to the claimant service

YetAnotherMatt:


src/web/routes/application/check/create-request-body.js
```diff
@@ -37,7 +38,8 @@ const createDeviceFingerprint = (headers) => ({
 
 const createRequestBody = (req) => ({
   claimant: createClaim(req.session.claim),
-  deviceFingerprint: createDeviceFingerprint(req.headers)
+  deviceFingerprint: createDeviceFingerprint(req.headers),
+  webUIVersion: APP_VERSION
```
gethyn1: Ideally `APP_VERSION` would be passed into the function rather than importing the config:
```
const createRequestBody = (config, req) => ({
  claimant: createClaim(req.session.claim),
  deviceFingerprint: createDeviceFingerprint(req.headers),
  webUIVersion: config.environment.APP_VERSION
})
```

YetAnotherMatt: done

PR #248, opened 2019-06-19
Feature/invalidate review state

gethyn1:
If a user has completed the application and changes their answer for 'Do you live in Scotland?' to 'yes', they should be shown the 'I live in Scotland' page rather than the 'check details' page.

This PR allows a step to invalidate the IN_REVIEW state based on the claim set in the session.


src/test/acceptance/features/do-you-live-in-scotland.feature
```diff
@@ -24,3 +24,10 @@ Feature: Do you live in Scotland?
     And I click continue
     Then I am shown the I live in Scotland page
     And I am informed that I cannot apply and should use the Scottish scheme
+
+  Scenario: I can change my answer to yes from the check details page
```
mikeSmith08: You should probably mention which question you're changing to yes in the Scenario title.

PR #253, opened 2019-06-20
HTBHF-1644 Change error message for failing to select Yes/No to be di…

mikeSmith08:
…fferent for the Do You Live In Scotland page to the Are You Pregnant page.


src/web/routes/application/are-you-pregnant/validate.js
```diff
@@ -50,7 +50,7 @@ const callValidateExpectedDeliveryDate = (req, res, next) =>
   check('expectedDeliveryDate').custom(validateExpectedDeliveryDate(res))(req, res, next)
 
 const validate = [
-  validateIsYesOrNo('areYouPregnant'),
+  validateIsYesOrNo('areYouPregnant', 'validation:selectYesOrNo'),
```
mikeSmith08: I'll leave this one until we know what the Are You Pregnant one should be, I don't think there's a story for it yet.

PR #254, opened 2019-06-20
HTBHF-1377 Added /agree-statement to the state machine

YetAnotherMatt:


YetAnotherMatt:
This has no functional effect yet, a larger PR will be coming up

src/web/routes/application/common/state-machine/state-machine.js
```diff
@@ -32,9 +32,9 @@ const stateMachine = {
     getNextAllowedPath: (req) => req.session.nextAllowedStep
   },
   [states.IN_REVIEW]: {
-    getNextPath: () => CHECK_URL,
+    getNextPath: (req) => req.path === CHECK_URL ? AGREE_STATEMENT_URL : CHECK_URL,
     isPathAllowed: (req, sequence) => isPathAllowed(sequence, req.session.nextAllowedStep, req.path),
-    getNextAllowedPath: (req) => req.session.nextAllowedStep,
+    getNextAllowedPath: (req) => req.path === AGREE_STATEMENT_URL ? AGREE_STATEMENT_URL : CHECK_URL,
```
mikeSmith08: Is this right, I thought it should still use `req.session.nextAllowedStep`, I'm obviously missing something here.

mikeSmith08:
As a general comment, would it be better to call the URL (and associated variables) `/agree-terms` rather than `/agree-statement`?

src/web/routes/application/common/state-machine/state-machine.js
gethyn1: This could be written as:
```
getNextAllowedPath: (req) => req.path === AGREE_STATEMENT_URL ? AGREE_STATEMENT_URL : req.session.nextAllowedStep
```
but we thought the other way was more explicit about what happens at this point i.e. the next path can only be `AGREE_STATEMENT_URL` or `CHECK_URL`

gethyn1: Yes you might be correct - I don't think this needs to change.

PR #255, opened 2019-06-20
HTBHF-1560 Content update to match latest prototype.

dwybourn:


src/web/routes/application/enter-dob/enter-dob.js
```diff
@@ -9,7 +9,8 @@ const pageContent = ({ translate }) => ({
   buttonText: translate('buttons:continue'),
   dayLabel: translate('enterDob.dayLabel'),
   monthLabel: translate('enterDob.monthLabel'),
-  yearLabel: translate('enterDob.yearLabel')
+  yearLabel: translate('enterDob.yearLabel'),
+  whyWeNeedThis: translate('enterDob.whyWeNeedThis')
```
dwybourn: Open to suggestions of better names.

mikeSmith08: What about `explanation` or `moreDetail`, I think either would be better.

YetAnotherMatt: explanation works for me

dwybourn: Updated to explanation.

src/web/views/enter-dob.njk
```diff
@@ -16,4 +18,9 @@
       text: hint
     }
   }) }}
+
+{{ govukInsetText({
+  text: explanation | safe
+}) }}
```
gethyn1: Does this need the `safe` filter?

dwybourn: No, removed it.

PR #257, opened 2019-06-21
HTBHF-1645 Updated content to match prototype.

dwybourn:


src/web/views/phone-number.njk
```diff
@@ -13,7 +13,7 @@
     }
   }) %}
 
-    <span class="govuk-hint govuk-!-margin-bottom-6">{{ formDescription }}</p>
+    <span class="govuk-hint govuk-!-margin-bottom-6">{{ formDescription }}</span>
```
gethyn1: As this is used in a number of places, should we create a Nunjucks macro, `macros/htbhf-hint.njk`?

src/web/views/enter-name.njk
```diff
@@ -12,7 +13,8 @@
     }
   }) %}
 
-    <p class="govuk-body">{{ formDescription }}</p>
+    <br>
```
gethyn1: Is this `<br />` for spacing? It would be better to use a class on the component above.

src/web/views/enter-name.njk
```diff
@@ -12,7 +13,8 @@
...
+    <span class="govuk-hint">{{ formDescription }}</span>
```
gethyn1: `{{ formDescription }}` doesn't seem like the correct variable name. Would `{{ hint }}` or `{{ nameFieldHint }}` be better?

src/web/views/phone-number.njk
dwybourn: Now using govukHint macro

src/web/views/enter-name.njk
dwybourn: Updated to hint.

src/web/views/enter-name.njk
dwybourn: Space no longer required so removed.

PR #263, opened 2019-06-24
Feature/scotland destroy session

mikeSmith08:
- Destroy the session when we go to the I Live In Scotland page adding optional middleware to register-form-routes and then only to the i-live-in-scotland page.
 - Modify the accessibility tests to deal with testing across two different sessions, do one test per session.
 - Add acceptance testing to verify that the session is destroyed (or not) at the correct place around the choice of Do You Live in Scotland page, this uses the value stored in the htbhf.sid cookie and the lang cookie.


src/test/a11y/test-suite.js
```diff
@@ -108,4 +100,32 @@ const runTests = async () => {
...
+}
+
+const runAllTests = async () => {
+  try {
+    let results = []
+    await runEndToEndTest(results)
+    await runILiveInScotlandTest(results)
+    handleTestResults(results)
+  } catch (error) {
+    console.log(error)
```
dwybourn: Should be console.error(), same for above.

mikeSmith08: Ok, will do, but that's what you put there in your original commit Dave. :-)

src/web/routes/application/register-form-routes.js
```diff
@@ -29,6 +29,7 @@ const createRoute = (config, csrfProtection, steps, router) => (step) =>
       configureGet(steps, step),
       getSessionDetails,
       handleRequestForPath(config, steps, step),
+      handleOptionalMiddleware(step.behaviour),
```
gethyn1: I wonder if this needs some reference to the request method? In the future we will need something similar for address lookup, but this will happen on `POST` rather than `GET`.

src/test/common/page/page.js
```diff
@@ -198,6 +200,35 @@ class Page {
...
   }
+
+  async getCurrentSessionId () {
+    const cookies = await this.getCookies()
+    const sessionCookie = cookies.find(this.getCookieByName('htbhf.sid'))
+    return (typeof sessionCookie !== 'undefined') ? sessionCookie.value : null
+  }
+
+  async getLangCookie () {
+    const cookies = await this.getCookies()
```
gethyn1: Should this line be in the `getCookieByName` method? This would stop it having to be defined in every method that aliases `getCookieByName` (`getCurrentSessionId` and `getLangCookie`).

mikeSmith08: Yep, good spot, I'll change.

src/web/routes/application/register-form-routes.js
mikeSmith08: I'm not sure what you mean, should this be calling a `handleOptionalGetMiddleware` method, is that what you mean, or do you mean it should be `step.getBehaviour`? If the latter, we'll have to be careful with the naming as this just looks like a getter.

gethyn1: I meant `step.getBehaviour`. Maybe it can stay as it is for now and we can reconsider when the time comes to implement something similar for `POST` requests.

YetAnotherMatt: `step.behaviourForGet`?

src/test/a11y/test-suite.js
```diff
@@ -34,19 +34,13 @@ const dateIn3Months = () => {
   Runs though the application, evaluating each page and performing post requests to populate the necessary
   data in the database.
  */
-const runTests = async () => {
+const runEndToEndTest = async (results) => {
```
YetAnotherMatt: @dwybourn this means the patch I sent you won't work without some changes

dwybourn: Ok. I'll apply and fix after this PR has gone in.

src/test/common/steps/overview-steps.js
```diff
@@ -17,3 +18,24 @@ Then(/^I am shown the overview page$/, async function () {
 Then(/^I am shown the first page of the application$/, async function () {
   await pages.waitForFirstPage()
 })
+
+// The original session id is stored on the Overview page, but to get a new session id we need to make sure
+// that we navigate to a subsequent page (back to the Overview page is sufficient) to check for a new session id
+// We also verify that the lang cookie has been removed
+Then(/^my session has been destroyed$/, async function () {
+  const pageSessionId = await pages.overview.sessionId
```
YetAnotherMatt: does this need an await - is getting overview.sessionId asynchronous? I expect it works iwth the await anyway...

mikeSmith08: That's true, it doesn't need it, it was there because that was originally retrieved using an asynchronous method, I'll correct it.

PR #269, opened 2019-06-25
Feature/integration tests

dwybourn:


src/test/a11y/test-suite.js
```diff
@@ -86,17 +86,16 @@ const runEndToEndTest = async (results) => {
...
     results.push(await pa11y(TERMS_AND_CONDITIONS_URL))
-    await postFormData(TERMS_AND_CONDITIONS_URL, {
-      ...formData,
-      'agree': 'agree'
-    }, requestCookie)
+    await postFormData(TERMS_AND_CONDITIONS_URL, { ...formData, agree: 'agree' }, requestCookie)
 
-    results.push(await pa11y(CONFIRM_URL))
+    const confirmResult = await pa11y(CONFIRM_URL)
+    if (confirmResult.documentTitle !== APPLICATION_COMPLETE_TITLE) {
```
YetAnotherMatt: Isn't this already handled by handleTestResults?

YetAnotherMatt: (Done by URL, rather than title). There's no harm in including this, though.

dwybourn: For the confirmation page we will have the same url for success and failure, it's only the title/heading that will change. So we need to check the title to make sure we got a successful response.

YetAnotherMatt: I had tested it with the claim service being down (which went back to the terms-and-conditions url), but not with a failure response.

PR #274, opened 2019-06-27
HTBHF-1563 updated are you pregnant messaging

YetAnotherMatt:


src/test/common/steps/are-you-pregnant.js
```diff
@@ -85,24 +85,24 @@ Then(/^I am informed that I need to select an option for are you pregnant$/, asy
...
 Then(/^I am informed that I need to enter an expected delivery date$/, async function () {
   await assertErrorHeaderTextPresent(pages.areYouPregnant)
-  await assertExpectedDeliveryDateErrorPresent('Enter your baby\'s due date')
+  await assertExpectedDeliveryDateErrorPresent('Enter the due date in the correct format')
 })
 
 Then(/^I am informed that the date is too far in the past$/, async function () {
-  await assertErrorHeaderTextPresent(pages.areYouPregnant, 'The due date you entered is more than 1 month ago')
-  await assertExpectedDeliveryDateErrorPresent('Please call our helpline on 0345 607 6823 to talk about your application.')
+  await assertErrorHeaderTextPresent(pages.areYouPregnant, 'There\'s a problem')
```
mikeSmith08: Shouldn't this be the other type of apostrophe? Same with below too.

src/web/server/locales/en/common.json
```diff
@@ -48,10 +48,11 @@
   "areYouPregnant": {
     "title": "Are you pregnant?",
     "heading": "Are you pregnant?",
-    "expectedDeliveryDateText": "When is your baby due to be born?",
+    "expectedDeliveryDateText": "When is your baby due? You must be at least 10 weeks pregnant.",
     "expectedDeliveryDateHint": "For example, {{exampleDate}}",
     "summaryKey": "Are you pregnant?",
-    "expectedDeliveryDateSummaryKey": "Baby's due date"
+    "expectedDeliveryDateSummaryKey": "Baby's due date",
```
mikeSmith08: Ok, I'm confused now on which apostrophes we're supposed to be using.

src/web/server/locales/en/validation.json
```diff
@@ -8,16 +8,15 @@
...
   "informationTooLong": "The information you entered is too long",
   "missingAddressField": "Tell us your address",
-  "expectedDeliveryDateInvalid": "Enter your baby's due date",
-  "expectedDeliveryDateTooFarInPastTitle": "The due date you entered is more than 1 month ago",
-  "expectedDeliveryDateTooFarInPast": "Please call our helpline on 0345 607 6823 to talk about your application.",
-  "expectedDeliveryDateTooFarInFutureTitle": "The date you have entered is more than 8 months in the future. You must be at least 10 weeks pregnant to apply for yourself.",
-  "expectedDeliveryDateTooFarInFuture": "If you have children under the age of 4, answer ‘no’ to this question to continue with this application on their behalf.",
+  "expectedDeliveryDateInvalid": "Enter the due date in the correct format",
+  "expectedDeliveryDateTooFarInPast": "Enter a due date to show you’re at least 10 weeks pregnant",
+  "expectedDeliveryDateTooFarInFuture": "Enter a due date to show you’re at least 10 weeks pregnant",
```
mikeSmith08: I don't understand how this will now work without changes to the njk file too, perhaps I'm missing something. I remember having to do this slightly differently for this scenario, hence why we needed the extra entries.

src/web/server/locales/en/validation.json
```diff
@@ -8,16 +8,15 @@
...
   "invalidPhoneNumber": "Enter in a UK mobile number",
```
mikeSmith08: It also looks like you'll need to do this to the Welsh file then too (presuming it's correct).

src/test/common/steps/are-you-pregnant.js
YetAnotherMatt: I thought I'd keep it the same as the other There's a problem headings, then change them all at the same time

src/web/server/locales/en/validation.json
YetAnotherMatt: I wasn't worried about the Welsh version, as it's not genuine text anyway

PR #277, opened 2019-06-28
HTBHF-1677 make childrens DOBs available to the view

gethyn1:
- Implement custom step behaviour for POST
- Add a single child's DOB to the session and make available to the view (on add) using custom step behaviour
- Initialise child count in session as a static number (1)

### DIFF
--- a/src/web/routes/application/children-dob/children-dob.js
+++ b/src/web/routes/application/children-dob/children-dob.js
@@ -16,6 +16,10 @@ const pageContent = ({ translate }) => ({
   aboutYourChild: translate('childrenDob.aboutYourChild')
 })
 
+const addActionRequested = body => body.hasOwnProperty('add')
+
+const extractChildrenEntries = (val, key) => key.startsWith('child')
+
 const behaviourForGet = (req, res, next) => {
   if (!req.session.hasOwnProperty('children')) {
     req.session.children = {}
@@ -27,10 +31,8 @@ const behaviourForGet = (req, res, next) => {
   next()
 }
 
-const extractChildrenEntries = (val, key) => key.startsWith('children')
-
 const behaviourForPost = (req, res, next) => {
-  if (req.body.hasOwnProperty('add')) {
+  if (addActionRequested(req.body)) {
     req.session.children = pickBy(extractChildrenEntries, req.body)
     return res.redirect(PATH)
   }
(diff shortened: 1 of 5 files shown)


src/web/routes/application/children-dob/children-dob.js
```diff
@@ -12,13 +16,39 @@ const pageContent = ({ translate }) => ({
...
+  req.session.children.count = 1
+
+  res.locals.children = req.session.children
+  next()
+}
+
+const extractChildrenEntries = (val, key) => key.startsWith('children')
+
+const behaviourForPost = (req, res, next) => {
+  if (req.body.hasOwnProperty('add')) {
```
mikeSmith08: It feels like we should have this in a function so its clearer what its doing, something like `addActionRequested`, but that's not quite right.

src/web/views/children-dob.njk
```diff
@@ -13,38 +13,17 @@
...
-
-    {% endcall %}
-
-    {{ htbhfDateInput({
-      fieldNamePrefix: "childrenDob",
-      claim: claim,
+  {%  for i in range(0, children.count) %}
+    {{ htbhfChildDobInput({
+      index: i + 1,
+      children: children,
```
mikeSmith08: I wonder if the input fields need to be `child` rather than `children`, `childrenDob-01` for example doesn't sound quite right, but `childDob-01` seems better.

src/web/routes/application/children-dob/children-dob.js
gethyn1: Agreed - I would like to add the actions to a map:
```
{
  add: () => {},
  remove: () => {}
}
```
.. but not sure that will work for this situation

src/web/routes/application/children-dob/children-dob.js
```diff
@@ -12,13 +16,39 @@ const pageContent = ({ translate }) => ({
   aboutYourChild: translate('childrenDob.aboutYourChild')
 })
 
+const behaviourForGet = (req, res, next) => {
+  if (!req.session.hasOwnProperty('children')) {
+    req.session.children = {}
+  }
+
+  req.session.children.count = 1
```
YetAnotherMatt: Shouldn't this be inside the if check? Otherwise won't it reset the count of children to 1 every get request?

src/web/views/children-dob.njk
YetAnotherMatt: Wouldn't it make sense to pass in the child (`children[i]`) to htbhfChildDobInput, rather than `children`?

YetAnotherMatt: Ignore me, I was forgetting what `children` was in this context

PR #278, opened 2019-07-01
Feature/add extra child

gethyn1:
- Increment children count in session on “add” action
- Use zero index for children count
- Add children to session on “submit”

### DIFF
--- a/src/web/routes/application/children-dob/children-dob.js
+++ b/src/web/routes/application/children-dob/children-dob.js
@@ -20,20 +20,31 @@ const addActionRequested = body => body.hasOwnProperty('add')
 
 const extractChildrenEntries = (val, key) => key.startsWith('child')
 
-const behaviourForGet = (req, res, next) => {
+const initialiseChildrenInSession = (req) => {
   if (!req.session.hasOwnProperty('children')) {
-    req.session.children = {}
+    req.session.children = {
+      count: 0
+    }
   }
 
-  req.session.children.count = 1
+  return req
+}
 
+const behaviourForGet = (req, res, next) => {
+  req = initialiseChildrenInSession(req)
   res.locals.children = req.session.children
   next()
 }
 
 const behaviourForPost = (req, res, next) => {
+  req = initialiseChildrenInSession(req)
+
   if (addActionRequested(req.body)) {
-    req.session.children = pickBy(extractChildrenEntries, req.body)
+    req.session.children = {
+      ...pickBy(extractChildrenEntries, req.body),
+      count: req.session.children.count += 1
+    }
+
     return res.redirect(PATH)
   }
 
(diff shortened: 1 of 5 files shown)


src/web/routes/application/children-dob/children-dob.js
```diff
@@ -20,20 +20,31 @@ const addActionRequested = body => body.hasOwnProperty('add')
...
   res.locals.children = req.session.children
   next()
 }
 
 const behaviourForPost = (req, res, next) => {
+  req = initialiseChildrenInSession(req)
+
+  req.session.children = {
+    ...pickBy(extractChildrenEntries, req.body),
+    count: req.session.children.count += 1
```
YetAnotherMatt: Should we always be adding a child on post? Or is this going to change as we add more functionality?

gethyn1: No .. good spot. We should update children on every POST, but only increment count on add.

src/web/views/children-dob.njk
```diff
@@ -13,7 +13,7 @@
     }
   }) %}
 
-  {%  for i in range(0, children.count) %}
+  {%  for i in range(0, children.count + 1) %}
```
YetAnotherMatt: Won't this always add one more child than is in the session?

src/web/routes/application/children-dob/children-dob.js
```diff
@@ -20,20 +20,34 @@ const addActionRequested = body => body.hasOwnProperty('add')
...
 
 const behaviourForPost = (req, res, next) => {
+  req = initialiseChildrenInSession(1, req)
+
+  const { count } = req.session.children
+
+  req.session.children = {
+    ...pickBy(extractChildrenEntries, req.body),
+    count
+  }
```
gethyn1: I'm starting to think this might be easier if we spilt `session.children` into:
```
{
  count: 1,
  entries: { ... }
}
```

src/web/views/children-dob.njk
gethyn1: Yes - it was working because on submit the count was not being incremented for the first child added, but that is wrong.

src/web/routes/application/children-dob/children-dob.js
```diff
@@ -20,20 +21,36 @@ const addActionRequested = body => body.hasOwnProperty('add')
...
+}
 
+const behaviourForGet = (req, res, next) => {
+  req = initialiseChildrenInSession(req)
   res.locals.children = req.session.children
   next()
 }
 
 const behaviourForPost = (req, res, next) => {
+  const childCount = countKeysContainingString('day', req.body)
```
YetAnotherMatt: Just a thought: is it worth matching on `-day-` to prevent accidentally matching on a word that happens to contain `day`?  I know there is no such input at the moment.

gethyn1: Good question. @mikeSmith08 any thoughts? Maybe we should be even more explicit and say `childDob-day-`

gethyn1: .. actually the format is `childDob-1-day` so would have to be `-day`

PR #280, opened 2019-07-01
HTBHF-1374 Added page to select how to receive your code.

dwybourn:


src/web/views/choose-channel-for-code.njk
```diff
@@ -0,0 +1,40 @@
...
+                text: text,
+                checked : claim.chooseChannelForCode === "text",
+                conditional: {
+                html: eddHtml
+            }
+            },
+            {
+                value: "email",
+                text: email,
+                checked : claim.areYouPregnant === "email"
```
YetAnotherMatt: `claim.areYouPregnant` => `claim.chooseChannelForCode`

YetAnotherMatt:
You don't seem to have included the changes to package.json? (I'm thinking of the change to compatibility tests to extend the timeout when running locally)

src/test/a11y/test-suite.js
```diff
@@ -22,6 +22,7 @@ const CARD_ADDRESS_URL = `${BASE_URL}/card-address`
 const PHONE_NUMBER_URL = `${BASE_URL}/phone-number`
 const EMAIL_ADDRESS_URL = `${BASE_URL}/email-address`
 const CHECK_URL = `${BASE_URL}/check`
+const CHOOSE_CHANNEL_FOR_CODE_URL = `${BASE_URL}/choose-channel-for-code`
```
gethyn1: Just checking this is the URL specified by UX?

dwybourn: That was choose-channel but Matt and I thought choose-channel-for-code would work better.

src/web/views/choose-channel-for-code.njk
dwybourn: Fixed.

src/test/a11y/test-suite.js
YetAnotherMatt: Something to double-check with Sam tomorrow...

package.json
```diff
@@ -23,7 +23,7 @@
     "test:smoke:debug": "mkdir -p build/reports && cross-env TESTS=smoke node $NODE_DEBUG_OPTION node_modules/cucumber/bin/cucumber-js src/test/smoke/features/ --require 'src/test/common/steps/*.js' --require 'src/test/smoke/config' -f json:build/reports/smoke-report.json",
     "test:smoke:report": "node src/test/smoke/smoke-report.js",
     "test:compatibility:local": "npm-run-all -s test:compatibility:local:test test:compatibility:local:report",
-    "test:compatibility:local:test": "mkdir -p build/reports/compatibility && cross-env TESTS=localCompatibility cucumber-js src/test/compatibility/features/ --require 'src/test/common/steps/*.js' --require 'src/test/compatibility/config/environment.js' -f json:build/reports/compatibility-report.json",
+    "test:compatibility:local:test": "mkdir -p build/reports/compatibility && cross-env TESTS=localCompatibility cucumber-js src/test/compatibility/features/ --require 'src/test/common/steps/*.js' --require 'src/test/acceptance/config/timeout.js' --require 'src/test/compatibility/config/environment.js' -f json:build/reports/compatibility-report.json",
```
mikeSmith08: I don't think we should be importing an acceptance test file into the compatibility tests, if we're going to reuse it, it should be moved out into the `common` package.

src/test/a11y/test-suite.js
mikeSmith08: My thoughts were that `channel` is a strange word to be using, but if they're happy with it, then I guess it's fine. My first thought would be to use the word `method` instead of `channel`.

package.json
gethyn1: I didn't see this when I approved somehow. What is going on here?

mikeSmith08:
I really wish you'd done this as a number of different PRs rather than one huge PR again, you could have easily at least done the acceptance tests as a separate PR.

package.json
YetAnotherMatt: Probably in common/driver then

src/test/common/steps/common-steps.js
```diff
@@ -109,6 +109,15 @@ async function selectNoOnChildrenThreeOrYoungerPage () {
   }
 }
 
+async function selectTextOnChooseChannelForCode () {
+  try {
+    await pages.chooseChannelForCode.selectTextRadioButton()
+    await pages.chooseChannelForCode.submitForm()
+  } catch (error) {
+    assert.fail(`Unexpected error caught trying to select 'Text' for 'We're going to send you the code' and submit the page - ${error}`)
```
mikeSmith08: Minor - its 'We're sending you a code'

YetAnotherMatt:
> I really wish you'd done this as a number of different PRs rather than one huge PR again, you could have easily at least done the acceptance tests as a separate PR.

We had to make some changes to the acceptance tests to get them to pass once the validation was in place. I think we could have added the page without any validation initially, however. That would have minimised the number of changes. Note for the future.

src/test/a11y/test-suite.js
YetAnotherMatt: or 'choose-where-we-send-code'...

src/test/common/steps/common-steps.js
dwybourn: Updated

package.json
dwybourn: Moved to common

package.json
```diff
@@ -14,16 +14,16 @@
...
-    "test:acceptance-no-wiremock": "mkdir -p build/reports && cross-env TESTS=acceptance cucumber-js src/test/acceptance/features/ --tags 'not @ignore' --tags 'not @RequiresWiremock' --require 'src/test/common/steps/*.js' --require 'src/test/acceptance/config' -f json:build/reports/acceptance-report.json",
-    "test:acceptance:debug": "mkdir -p build/reports && cross-env TESTS=acceptance node $NODE_DEBUG_OPTION node_modules/cucumber/bin/cucumber-js src/test/acceptance/features/ --tags 'not @ignore' --require 'src/test/common/steps/*.js' --require 'src/test/acceptance/config' -f json:build/reports/acceptance-report.json",
+    "test:acceptance": "mkdir -p build/reports && cross-env TESTS=acceptance cucumber-js src/test/acceptance/features/navigation.feature --tags 'not @ignore' --require 'src/test/common/steps/*.js' --require 'src/test/acceptance/config' --require 'src/test/common/timeout.js' -f json:build/reports/acceptance-report.json",
```
YetAnotherMatt: `navigation.feature` has to come out again - that was there for local testing. 
Rather than making this same change in several lines here, it might be better if there was a script in the config directory that imported the script from common.

gethyn1: should this have `navigation.feature` in the command?

dwybourn: Removed.

dwybourn: Ok. I'll add that to the next PR so this one can get through without an extra file and a rebuild.

YetAnotherMatt: I'd get other opinions on that before actually making changes. There's an argument that making it explicit is good - it took me a while to figure how how it was being executed for the acceptance tests.

PR #281, opened 2019-07-01
HTBHF-1374 Acceptance tests for choose channel for code

dwybourn:


src/test/acceptance/features/choose-channel-for-code.feature
```diff
@@ -0,0 +1,25 @@
...
+  Scenario: Text and Email are displayed
+    Then Text and Email options are displayed on the choose channel for code page
+
+  Scenario: Do not select an option
+    When I do not select an option
+    And I click continue
+    Then I am informed that I need to select an option for choose channel for code
+
+  Scenario: Select the Text option
+    When I select Text to how would I like to receive my code
```
YetAnotherMatt: Minor suggested wording change: `When I answer 'Text' to how would I like to receive my code`

mikeSmith08: Or maybe `When I choose Text as the method to receive the code`. I do like the word method rather than channel for this page, I think we should run this past Sam before we go too far with this story and have loads of renaming to do.

src/test/common/page/choose-channel-for-code.js
```diff
@@ -37,6 +37,10 @@ class ChooseChannelForCode extends SubmittablePageWithRadioButtons {
   async selectTextRadioButton () {
     await super.selectRadioButton('text')
   }
+
+  async selectEmailRadioButton () {
+    await super.selectRadioButton('email')
```
mikeSmith08: I think `email` and `text` should be constants. I was going to say that on the previous PR but you merged it before I could review it!

src/test/common/steps/choose-channel-for-code-steps.js
```diff
@@ -1,7 +1,37 @@
...
 
 Then(/^I am shown the choose channel for code page$/, async function () {
   await pages.chooseChannelForCode.waitForPageLoad()
 })
+
+Then(/^Text and Email options are displayed on the choose channel for code page$/, async function () {
+  const labels = await pages.chooseChannelForCode.getAllRadioLabels()
+  const text = await Promise.all(labels.map(async (label) => label.getText()))
+
+  expect(text).to.include('Text')
```
mikeSmith08: These two should be constants too.

src/test/acceptance/features/choose-channel-for-code.feature
dwybourn: Went for `I select Text as the method to receive the code` as the use of `I select` is used for both are you pregnant and do you live in Scotland.

src/test/common/page/choose-channel-for-code.js
dwybourn: Constants added.

src/test/common/steps/choose-channel-for-code-steps.js
dwybourn: Constants added.

PR #283, opened 2019-07-02
HTBHF-1683 Fixed back link behaviour when applicant has no children

YetAnotherMatt:


src/test/common/steps/add-children-dob-steps.js
```diff
@@ -1,7 +1,12 @@
...
+const { assertBackLinkUrlIsEqualTo } = require('./common-assertions')
 
 const pages = require('./pages')
 
 Then(/^I am shown the add your childrens dates of birth page$/, async function () {
   await pages.addChildrenDOB.waitForPageLoad()
 })
+
+Then(/^The back link points to the Add your children’s dates of birth page$/, async function () {
+  await assertBackLinkUrlIsEqualTo(`${pages.url}${pages.addChildrenDOB.getPath()}`)
```
mikeSmith08: I'm wondering whether this should imply take in the expected page and then we can do the building and checking for the expected URL more generically in Page rather than like this. I'm not sure if I'm describing that well, happy to talk it through.

YetAnotherMatt: done

PR #291, opened 2019-07-03
Childrens‘ date of birth validation

gethyn1:
- Fix error message in child dob macro
- Add validation translations for child date of birth
- Add validator for date more than four months in the past
- Add extra validation for childrens‘ date of birth:
1. date more than four years in the past
2. name too long

### DIFF
--- a/src/web/server/locales/en/validation.json
+++ b/src/web/server/locales/en/validation.json
@@ -21,6 +21,8 @@
   "invalidEmail": "Enter an email address in the correct format, like name@example.com",
   "acceptTermsAndConditions": "Confirm that you’ve read and will comply with these terms and conditions",
   "selectTextOrEmailChannel": "Select if you want to receive a code by text or email",
-  "childsDateOfBirthInvalid": "Enter your child’s date of birth in the correct format",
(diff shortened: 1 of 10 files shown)


src/web/server/locales/en/validation.json
```diff
@@ -21,6 +21,8 @@
   "invalidEmail": "Enter an email address in the correct format, like name@example.com",
   "acceptTermsAndConditions": "Confirm that you’ve read and will comply with these terms and conditions",
   "selectTextOrEmailChannel": "Select if you want to receive a code by text or email",
-  "childsDateOfBirthInvalid": "Enter your child’s date of birth in the correct format",
-  "enterTheSixDigitCodeWeSentYou": "Enter the 6 digit code we sent you"
+  "enterTheSixDigitCodeWeSentYou": "Enter the 6 digit code we sent you",
+  "childDateOfBirthInvalid": "Enter your child’s date of birth in the correct format",
```
mikeSmith08: Was Sam happy with this error message, I still think the use of `in the correct format` seems odd? If so, are we going to change the enter DOB and expected delivery date ones at some point to conform?

PR #296, opened 2019-07-04
HTBHF-1738 Updated accessibility tests to obtain 2fa confirmation code from the session details app

YetAnotherMatt:


src/web/routes/application/send-code/send-code.js
```diff
@@ -10,11 +11,17 @@ const pageContent = ({ translate }) => ({
   explanation: translate('sendCode.explanation')
 })
 
+const behaviourForPost = (req, res, next) => {
+  req.session[CONFIRMATION_CODE_SESSION_PROPERTY] = '123456'
```
YetAnotherMatt: I've not written any tests for this as it will change soon to use a unique code, not a constant

PR #298, opened 2019-07-04
Feature/notify send code

dwybourn:
Send request to notify to either email or text the user a confirmation code depending on the option they selected. Confirmation code is currently hardcoded at '123456' and will be randomly generated in a future PR.


src/web/routes/application/send-code/notify.js
```diff
@@ -0,0 +1,29 @@
...
+const notifyClient = new NotifyClient(config.environment.NOTIFY_API_KEY)
+require('dotenv').config()
+
+const { EMAIL, TEXT } = require('../common/constants')
+
+const EMAIL_TEMPLATE_ID = '18cf4f69-e1b5-4aa9-bed7-c906d3a59285'
+const SMS_TEMPLATE_ID = 'a770479a-097c-4079-acc2-d452a6f21585'
+
+function sendConfirmationCode (claim, channelForCode, confirmationCode) {
+  const notifyOptions = { personalisation: { confirmationCode }, reference: uuid() }
```
YetAnotherMatt: I suggest we extract the reference and include it in a log entry for debugging

package.json
```diff
@@ -102,11 +102,13 @@
     "libphonenumber-js": "^1.7.19",
     "moment": "^2.24.0",
     "node-sass": "^4.11.0",
+    "notifications-node-client": "^4.6.0",
     "nunjucks": "^3.2.0",
     "ramda": "^0.26.1",
     "redis": "^2.8.0",
     "request": "^2.88.0",
     "request-promise": "^4.2.2",
+    "uuid": "^3.3.2",
```
gethyn1: Do we need both `uuidv4` and `uuid`?

src/config/vcap-services.js
```diff
@@ -9,17 +9,25 @@ function getVCAPServices (service) {
   return services
 }
 
-function getVariableService (userProvidedServices) {
-  const variableServices = userProvidedServices.filter(service => service.instance_name === 'variable-service')
+function getService (userProvidedServices, serviceName) {
+  const variableServices = userProvidedServices.filter(service => service.instance_name === serviceName)
```
gethyn1: Should this variable just be `services` to match the updated function name?

package.json
dwybourn: Didn't noticed we already had uuidv4, I'll remove uuid.

src/web/routes/application/send-code/notify.js
```diff
@@ -0,0 +1,29 @@
...
+const config = require('../../../../config')
+const notifyClient = new NotifyClient(config.environment.NOTIFY_API_KEY)
+require('dotenv').config()
+
+const { EMAIL, TEXT } = require('../common/constants')
+
+const EMAIL_TEMPLATE_ID = '18cf4f69-e1b5-4aa9-bed7-c906d3a59285'
+const SMS_TEMPLATE_ID = 'a770479a-097c-4079-acc2-d452a6f21585'
+
+function sendConfirmationCode (claim, channelForCode, confirmationCode) {
```
gethyn1: Could the arguments for this just be
```
sendConfirmationCode (claim, channel, code)
```
as the extra words seem to be implied by the function name?

src/config/vcap-services.js
dwybourn: Updated to services.

src/web/routes/application/send-code/notify.js
dwybourn: Done.

src/web/routes/application/send-code/notify.js
```diff
@@ -0,0 +1,29 @@
+const { NotifyClient } = require('notifications-node-client')
+const uuid = require('uuid/v4')
+const config = require('../../../../config')
+const notifyClient = new NotifyClient(config.environment.NOTIFY_API_KEY)
```
gethyn1: Ideally `config` would be passed in here rather than imported directly, although I realise this could complicate things. One approach would be:

```
// Initialise the client in behaviourForPost()
const behaviourForPost = (config) => {
  const notifyClient = new NotifyClient(config.environment.NOTIFY_API_KEY) 
  const sendConfirmationCodeWithNotify = sendConfirmationCode(notifyClient) 
  
  return (req, res, next) => {
    ...
    sendConfirmationCodeWithNotify(claim, channel, code)
    ...
  }
}

// Pass the client to sendConfirmationCode()
const sendConfirmationCode = (client) => (claim, channel, code) => { ... }
```

src/web/routes/application/send-code/notify.js
dwybourn: Updated with simpler names.

src/config/vcap-services.js
```diff
@@ -9,17 +9,25 @@ function getVCAPServices (service) {
...
 }
 
-function getVariableService (userProvidedServices) {
-  const variableServices = userProvidedServices.filter(service => service.instance_name === 'variable-service')
-  if (variableServices.length !== 1) {
-    throw new Error(`Expected exactly one variable-service, instead found ${variableServices.length}`)
+function getService (userProvidedServices, serviceName) {
+  const services = userProvidedServices.filter(service => service.instance_name === serviceName)
+  if (services.length !== 1) {
+    throw new Error(`Expected exactly one variable-service, instead found ${services.length}`)
```
mikeSmith08: You should change the reference to the `variable-service` in the message too.

mikeSmith08:
It would be useful on a big PR like this to have a comment describing what the PR is for rather than having to figure it out from the contents of the PR as you go along, especially when you just give the pr the title `feature/notify send code` - that has been what you've been working on a while so its to vague. You need to get better at this kind of thing!

src/web/routes/application/send-code/notify.js
```diff
@@ -0,0 +1,32 @@
...
+const uuid = require('uuidv4')
+const config = require('../../../../config')
+const notifyClient = new NotifyClient(config.environment.NOTIFY_API_KEY)
+const { logger } = require('../../../logger')
+require('dotenv').config()
+
+const { EMAIL, TEXT } = require('../common/constants')
+
+const EMAIL_TEMPLATE_ID = '18cf4f69-e1b5-4aa9-bed7-c906d3a59285'
+const SMS_TEMPLATE_ID = 'a770479a-097c-4079-acc2-d452a6f21585'
```
mikeSmith08: Where do these values come from, some documentation would be good here, how do I know if they're right or not when reviewing?

src/web/routes/application/send-code/notify.js
```diff
@@ -0,0 +1,32 @@
...
+
+function sendConfirmationCode (claim, channel, code) {
+  const reference = uuid()
```
mikeSmith08: What is this used for, why is it a uuid?

src/web/routes/application/send-code/notify.js
mikeSmith08: Presumably these templates are specific to the 2FA process, so might be better to rename these constants so they are specific to that process.

src/web/routes/application/send-code/send-code.js
```diff
@@ -12,7 +15,13 @@ const pageContent = ({ translate }) => ({
 })
 
 const behaviourForPost = (req, res, next) => {
-  req.session[CONFIRMATION_CODE_SESSION_PROPERTY] = '123456'
+  if (validationResult(req).isEmpty()) {
+    // TODO DW HTBHF-1735 Randomly generate the code
+    const confirmationCode = '123456'
+    sendConfirmationCode(req.session.claim, req.body.channelForCode, confirmationCode)
+    req.session[CONFIRMATION_CODE_SESSION_PROPERTY] = '123456'
```
mikeSmith08: I presume you should be using `confirmationCode` here.

sample.env
```diff
@@ -20,3 +20,4 @@ MAINTENANCE_MODE=false
 # date format must be HH:mm DD/MM/YYYY, only used if MAINTENANCE_MODE is true
 SERVICE_AVAILABLE_DATE=14:30 02/03/2019
 APP_VERSION=local
+NOTIFY_API_KEY=notify-test-key-replace-with-real-one-in-1password
```
gethyn1: Should we be saying `1password` here as this is publicly available? Maybe just use `xxxxxxx-xxxxxxx-xxxxx-xxxxx` or similar

src/config/vcap-services.js
```diff
@@ -9,17 +9,25 @@ function getVCAPServices (service) {
...
   }
-  return variableServices[0]
+  return services[0]
 }
 
 const getVariableServiceCredentials = () => {
+  return getServiceCredentials('variable-service')
```
gethyn1: Not totally necessary but this could be written without the `return`:
```
const getVariableServiceCredentials = () => getServiceCredentials('variable-service')
```

src/test/session-details-provider/session-details-app.js
```diff
@@ -38,3 +38,10 @@ process.on('SIGTERM', () => {
     process.exit(0)
   })
 })
+
+process.on('SIGINT', () => {
```
gethyn1: I think this should be in a separate PR. It's quite a complicated PR anyway and this adds extra mental overhead.

src/web/routes/application/send-code/send-code.js
dwybourn: That constant is for `confirmationCode`.

sample.env
dwybourn: I'll remove the 1password section but it's not really a security concern as people would still need to know our 1password credentials.

src/test/session-details-provider/session-details-app.js
dwybourn: Removed, will update in it's own PR.

src/config/vcap-services.js
dwybourn: Missed this when I was refactoring. Removed the extra functions which allowed the test class to be smaller as well.

src/web/routes/application/send-code/notify.js
dwybourn: Add a comment explaining.

src/web/routes/application/send-code/notify.js
dwybourn: Added a comment stating that these match the templates in govuk, but won't rename as the context of send-code implies 2FA.

src/config/vcap-services.js
dwybourn: Updated error message.

src/web/routes/application/send-code/notify.js
dwybourn: Unfortunately this doesn't seem to be that simple due to changes in register-form-routes. Happy to pair with it in the next PR as doing it in this PR would require an additional file change.

README.md
```diff
@@ -7,6 +7,7 @@
 
 ## Environment variables
 Before starting the application or running tests, ensure that the correct environment variables have been set. A list of required environment variables are available in `sample.env`.
+The environment variable `NOTIFY_API_KEY` is required and needs to be a real test api key (not the whitelisted team one).
```
gethyn1: I find this line a bit confusing. Why is this a test api key?

dwybourn: I'll clarify, locally in .env you need to use a test one.

src/web/routes/application/send-code/notify.js
gethyn1: OK, for the sake of 2fa this is fine for now. I would really like to come back to this so will have a think about it. I think maybe the solution would be to pass `config, steps, step` to all middleware in `registerFormRoutes` as generally most steps end up needing these parameters at some point.

src/web/routes/application/send-code/notify.js
```diff
@@ -0,0 +1,34 @@
+const { NotifyClient } = require('notifications-node-client')
+const uuid = require('uuidv4')
+const config = require('../../../../config')
+const notifyClient = new NotifyClient(config.environment.NOTIFY_API_KEY)
+const { logger } = require('../../../logger')
+require('dotenv').config()
```
gethyn1: I don't think you need `dotenv` as you're not accessing a variable on `process.env`

dwybourn: Removed.

PR #300, opened 2019-07-04
HTBHF-1738 Added a work-around for iPhone 8 to get the confirmation code in the browser instead of a separate get request

YetAnotherMatt:


mikeSmith08:
I'm not sure I understand why we need this and what problem it fixes, can you explain please?

dwybourn:
I'm a bit confused, surely if there is no session id something has gone wrong in the test and we should it fail?

PR #303, opened 2019-07-05
HTBHF-1678 Add validation failure scenarios for Add Child DOB page, plus test for 10 children.

mikeSmith08:
- I've also made some changes around the wording of steps and the methods they use to prefer the word 'submit' over 'enter' when those steps enter the details and also click the continue button as a part of the one step.
- I'm also aware that adding the test for 10 children isn't a failure scenario and could possibly have been in a subsequent PR, I'm happy to remove it if people feel strongly enough.
- I've also added the ability to add a day increment to the date creation method so that the children can each have different dates of birth based on the current index value.
- The majority of the methods now allow providing a child name, but also have a default if the test doesn't care (which is also index based).

### DIFF
--- a/src/test/common/steps/add-children-dob-steps.js
+++ b/src/test/common/steps/add-children-dob-steps.js
@@ -1,21 +1,42 @@
 const { When, Then } = require('cucumber')
 const { assert } = require('chai')
-const { assertBackLinkPointsToPage } = require('./common-assertions')
-const { enterChildUnder3Details } = require('./common-steps')
+const { assertBackLinkPointsToPage, assertErrorHeaderTextPresent, assertFieldErrorAndLinkTextPresentAndCorrect } = require('./common-assertions')
+const { submitChild3OrUnderDetails } = require('./common-steps')
+const { LONG_STRING } = require('./constants')
 
 const pages = require('./pages')
 
 When(/^I submit the details of my child who is three or younger$/, async function () {
-  await enterChildUnder3Details()
+  await submitChild3OrUnderDetails()
   await pages.genericPage.submitForm()
 })
 
-When(/^I enter the details of my child who is three or younger without a name$/, async function () {
-  await enterChildUnder3DetailsWithoutAName()
+When(/^I submit the details of my child who is three or younger without a name$/, async function () {
+  await submitChild3OrUnderDetails('')
 })
 
-When(/^I enter the details of my two children who are three or younger/, async function () {
-  await enterTwoSetsOfChildrensDatesOfBirth()
+When(/^I submit the details of my child who is three or under with a very long name$/, async function () {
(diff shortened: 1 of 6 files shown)


src/test/common/steps/add-children-dob-steps.js
```diff
@@ -26,22 +47,74 @@ Then(/^The back link points to the Add your children’s dates of birth page$/,
...
     await pages.addChildrenDOB.submitForm()
   } catch (error) {
-    assert.fail(`Unexpected error caught trying to enter child details and submit the page - ${error}`)
+    assert.fail(`Unexpected error caught trying to enter ten sets of children details and submit the page - ${error}`)
+  }
+}
+
+async function enterChildDetailsAndClickAddAnother (childIndex) {
+  await pages.addChildrenDOB.enterChild3OrUnderDetails(`Child${childIndex}`, childIndex)
+  if (childIndex < 10) {
```
dwybourn: Do we need this check here, feels like the for loop it's called from is enough to suffice there being no more than ten children.

mikeSmith08: This is here to make sure that we don't press the Add Another Child on the final one otherwise we get a validation error on the page on submit due to the final one being empty.

gethyn1: I agree with @dwybourn - this function has to have knowledge of the function that is calling it. Could this be handled with a second parameter, something like:
```
enterChildDetailsAndClickAddAnother (childIndex, shouldAdd = true) { ... }

enterChildDetailsAndClickAddAnother(childIndex, childIndex !== 10)
```

src/test/common/steps/add-children-dob-steps.js
```diff
@@ -26,22 +47,74 @@ Then(/^The back link points to the Add your children’s dates of birth page$/,
...
   } catch (error) {
     assert.fail(`Unexpected error caught trying to enter two sets of children details and submit the page - ${error}`)
   }
 }
 
-async function enterChildUnder3DetailsWithoutAName () {
+async function submitTenSetsOfChildren3OrUnderDetails () {
   try {
-    await pages.addChildrenDOB.enterChildUnder3DetailsWithoutAName()
+    var childIndex
```
gethyn1: This `var` can be inlined:
```
for (let childIndex = 1; childIndex <= 10; childIndex++) { ... }
```

mikeSmith08: Thanks, I've updated.

src/test/common/steps/add-children-dob-steps.js
mikeSmith08: Yes, good point @gethyn1, I've updated as you have described.

PR #310, opened 2019-07-08
HTBHF-1754 Updated enter code and send code to to be navigable once t…

dwybourn:
Enter code and send code can no longer be navigated to once the code has been entered. 

This on it's own would break the back link behaviour of check details so this has been updated to get the last navigable step (email address) instead of the last step in the application.

Check detail test will be added in next PR.

A future story handles the behaviour of being able to re-enter the code if the user changes their email/phone number.


src/web/routes/application/enter-code/enter-code.js
```diff
@@ -28,13 +31,26 @@ const behaviourForGet = (req, res, next) => {
   next()
 }
 
+const behaviourForPost = (req, reqs, next) => {
+  if (validationResult(req).isEmpty()) {
+    req.session[CONFIRMATION_CODE_ENTERED_SESSION_PROPERTY] = true
```
YetAnotherMatt: I think in later PR's we'll want to replace this (or add to it) with information about which email address/phone number was validated.

dwybourn: We could log that information, not sure how much value it has in the session

src/test/acceptance/features/navigation.feature
```diff
@@ -60,7 +60,7 @@ Feature: Application process navigation is controlled
       | enter date of birth | check details   |
       | are you pregnant    | confirmation    |
 
-  Scenario Outline: Navigation to previous steps in an unsubmitted application flow is allowed
+  Scenario Outline: Navigation to previous steps (excluding enter code and send code) in an unsubmitted application flow is allowed
```
gethyn1: Could this just be:
```
Navigation to previous navigable steps in an unsubmitted application flow is allowed
```

src/web/routes/application/check/get.js
```diff
@@ -47,13 +40,12 @@ const getCheck = (steps) => (req, res) => {
     claim: req.session.claim,
     ...pageContent({ translate: req.t }),
     checkRowData: getFlattenedRowData(req)(steps),
-    previous: getLastStepPath(steps)
+    previous: getPreviousPath(steps, steps[steps.length - 1], req.session)
```
gethyn1: Is this going to work? If the last step's `isNavigable()` function returns true then it looks like this would return the path for the second last step, which would be incorrect.

dwybourn: You;re right, it's only working now because we want to skip back over two steps.

src/web/routes/application/enter-code/enter-code.js
YetAnotherMatt: I think we may need it when deciding whether we need to send a code out again (they've gone back several steps, and clicked continue through the email and mobile number pages again, but did they make changes?). We may also need to store that information in the db with the claim (I'm not sure of the need for this yet). Either way, it's not required for this PR.

src/web/routes/application/check/get.js
dwybourn: Updated to get the last step if navigable, otherwise call getPreviousPath

src/test/acceptance/features/navigation.feature
dwybourn: Updated

src/web/routes/application/enter-code/enter-code.js
dwybourn: We do have a story to cover the user changing their email/phone number. There we would invalidate the sending of the code if they changed the channel used for the code.

src/web/routes/application/check/get.js
```diff
@@ -38,6 +31,17 @@ const getRowData = (req) => (step) => {
   return Array.isArray(result) ? result.map(applyPathToRow) : applyPathToRow(result)
 }
 
+// a step is navigable if it hasn't defined an isNavigable function.
+const stepIsNavigable = (lastStep, session) => !lastStep.hasOwnProperty('isNavigable') || lastStep.isNavigable(session)
```
gethyn1: I would change `lastStep` -> `step` in the parameters as this function doesn't really have a concept of `last`

dwybourn: Updated to step.

src/web/routes/application/check/get.js
```diff
@@ -38,6 +31,17 @@ const getRowData = (req) => (step) => {
   return Array.isArray(result) ? result.map(applyPathToRow) : applyPathToRow(result)
 }
 
+// a step is navigable if it hasn't defined an isNavigable function.
+const stepIsNavigable = (step, session) => !step.hasOwnProperty('isNavigable') || step.isNavigable(session)
+
+const getLastNavigablePath = (steps, req) => {
```
gethyn1: Is there a missing unit test for `getLastNavigablePath`?

src/web/public/styles/all.css
```diff
@@ -0,0 +1 @@
...set:3px;box-shadow:0 0 0 4px #ffbf47}.govuk-radios__input:checked+.govuk-radios__label::after{opacity:1}.govuk-radios__input:disabled,.govuk-radios__input:disabled+.govuk-radios__label{cursor:default}.govuk-radios__input:disabled+.govuk-radios__label{opacity:.5}@media (min-width: 40.0625em){.govuk-radios--inline:after{content:"";display:block;clear:both}.govuk-radios--inline .govuk-radios__item{margin-right:20px;float:left;clear:none}}.govuk-radios--inline.govuk-radios--conditional .govuk-radios__item{margin-right:0;float:none}.govuk-radios__divider{font-family:"nta",Arial,sans-serif;-webkit-font-smoothing:antialiased;-moz-osx-font-smoothing:grayscale;font-weight:400;font-size:16px;font-size:1rem;line-height:1.25;color:#0b0c0c;width:40px;margin-bottom:10px;text-align:center}@media print{.govuk-radios__divider{font-family:sans-serif}}@media (min-width: 40.0625em){.govuk-radios__divider{font-size:19px;font-size:1.1875rem;line-height:1.31579}}@media print{.govuk-radios__divider{font-size:14pt;line-height:1.15}}@media print{.govuk-radios__divider{color:#000}}.govuk-radios__conditional{margin-bottom:15px;margin-left:18px;padding-left:33px;border-left:4px solid #bfc1c3}@media (min-width:
```
mikeSmith08: Why is this committed, shouldn't it be ignored?

src/web/routes/application/check/get.js
dwybourn: Added unit test.

src/web/public/styles/all.css
dwybourn: There's a missing '/' at the end of the rule which we missed during the code review. Can update it in the PR?

src/web/routes/application/check/get.test.js
```diff
@@ -1,5 +1,41 @@
...
-const { getRowData, getLastStepPath, getFlattenedRowData } = require('./get')
+const proxyquire = require('proxyquire')
+const sinon = require('sinon')
+
+const { getRowData, getFlattenedRowData } = require('./get')
+
+const getPreviousPath = sinon.spy()
+const { getLastNavigablePath } = proxyquire('./get', { '../common/get-previous-path': { getPreviousPath } })
+
+test('getLastNavigablePath returns path of last step if last step has not isNavigable function', (t) => {
```
gethyn1: Typo `... if last step has not isNavigable ...` -> `... if last step has no isNavigable ...`

src/web/routes/application/check/get.test.js
```diff
@@ -1,5 +1,41 @@
...
+  const steps = [{ path: '/first' }, { path: '/last' }]
+  const session = {}
+
+  const result = getLastNavigablePath(steps, session)
+
+  t.equal(result, '/last')
+  t.end()
+})
+
+test('getLastNavigablePath returns path of last step if last step isNavigable return true', (t) => {
```
gethyn1: This doesn't quite make sense, I think should be something like `returns path of last step if last step's isNavigable function returns true`

src/web/routes/application/check/get.test.js
```diff
@@ -68,7 +80,11 @@ test('getFlattenedRowData returns flattened row data', (t) => {
   const result = getFlattenedRowData(req)(steps)
 
   t.deepEqual(result,
-    [{ keyA: 'myKey1', valueA: 'myValue1', path: 'mypath1' }, { keyB: 'myKey3', valueB: 'myValue3', path: 'mypath1' }, { key2: 'myKey2', value: 'myValue2', path: 'mypath2' }],
+    [{ keyA: 'myKey1', valueA: 'myValue1', path: 'mypath1' }, { keyB: 'myKey3', valueB: 'myValue3', path: 'mypath1' }, {
```
gethyn1: Is there a reason for this reformatting?

dwybourn: It was done by webstorm, I'll put it back

src/web/routes/application/check/get.test.js
dwybourn: Fixed typo

src/web/routes/application/check/get.test.js
dwybourn: Updated.

src/web/routes/application/check/get.test.js
dwybourn: Webstorm reformatted because it's over 160 lines, I've put it back but we might want to add a max line length?

src/web/routes/application/check/get.test.js
dwybourn: Fixed typo.

PR #312, opened 2019-07-08
HTBHF-1685 Add children's dates of birth to the claim being posted to the claimant service.

mikeSmith08:


src/web/routes/application/terms-and-conditions/create-request-body.js
```diff
@@ -12,7 +12,20 @@ const createExpectedDeliveryDate = (claim) => {
...
 }
 
-const createClaim = (claim) => ({
+const createChildrenDobArray = (children) => {
+  if (typeof children === 'undefined') {
+    return null
+  }
+
+  let childrenArray = []
+  for (let i = 1; i <= children.childCount; i++) {
```
dwybourn: Can we add a comment to explain why we're starting at 1 instead of 0.

mikeSmith08: We have to do it this way because on the page the inputs are stored as child1, child2, so this has to match. Here's an example of how the data is stored in the session:
`
{
  "childName-1": "First",
  "childDob-1-day": "1",
  "childDob-1-month": "1",
  "childDob-1-year": "2018",
  "childName-2": "Second",
  "childDob-2-day": "2",
  "childDob-2-month": "2",
  "childDob-2-year": "2017",
  "childDob-1": "2018-01-01",
  "childDob-2": "2017-02-02",
  "inputCount": 2,
  "childCount": 2
}`

PR #314, opened 2019-07-09
Feature/navigation tests

dwybourn:
Added acceptance tests to confirm that users can not navigate to send-code or enter-code after the user has entered their confirmation code.

Added acceptance test for back link behaviour of check details page. Clicking the back link should go to the last navigable page (currently email address), skipping enter code and send code as they are not navigable after the confirmation code has been entered.


src/test/common/steps/check-details.js
```diff
@@ -68,6 +68,11 @@ When(/^I choose to change my answer to Do you live in Scotland$/, async function
   await pages.check.clickChangeLinkFor('Do you live in Scotland?')
 })
 
+When(/^I click on the back link$/, async function () {
```
mikeSmith08: This is inconsistent with how we've done it on other stories, we have a function in `common-assertions.js` called `assertBackLinkPointsToPage` which we should probably use instead of this. I also think that a step like this should probably be in `common-steps.js` too, and should be calling a function in `Page.js` rather than something specific to the Check Details page.

src/test/common/steps/navigation.js
```diff
@@ -175,3 +175,11 @@ When(/^I click the Cookies link$/, async function () {
 When(/^I navigate to the (.*) page$/, async function (page) {
   await navigateToPage(page)
 })
+
+When(/^I go directly to the send code page$/, async function () {
```
mikeSmith08: You don't need to write these steps, you can use the `I navigate to the XXXX page` and as long as you've added the page to the list in `pages.js` and the page name matches what you have in your `getPageName()` function matching this then it'll all work for you.

dwybourn: That won't work as that will go to the page and wait for that page to load, which it never will as we're being redirected to a different page.

src/test/common/steps/check-details.js
dwybourn: Updated to now just assert that the back link url is correct using the `assertBackLinkPointsToPage` method.

src/test/common/steps/navigation.js
dwybourn: As discussed, updated to use openDirect.

PR #316, opened 2019-07-09
HTBHF-1702 Now validating that the confirmation code added is correct…

dwybourn:
Now generating a random six digit code for the confirmation code.
Now validating that the user has entered the correct code.

### DIFF
--- a/src/web/routes/application/enter-code/validate.js
+++ b/src/web/routes/application/enter-code/validate.js
@@ -1,9 +1,11 @@
+const { CONFIRMATION_CODE_SESSION_PROPERTY } = require('../common/constants')
 const { check } = require('express-validator')
 const { translateValidationMessage } = require('../common/translate-validation-message')
 
+const validateConfirmationCodeIsCorrect = () => (confirmationCode, { req }) => req.body.confirmationCode === req.session[CONFIRMATION_CODE_SESSION_PROPERTY]
+
 const validate = [
-  check('confirmationCode')
-    .not().isEmpty()
+  check('confirmationCode').custom(validateConfirmationCodeIsCorrect())
     .withMessage(translateValidationMessage('validation:enterTheSixDigitCodeWeSentYou'))
 ]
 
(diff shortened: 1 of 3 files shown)


src/web/routes/application/enter-code/validate.js
```diff
@@ -1,9 +1,11 @@
+const { CONFIRMATION_CODE_SESSION_PROPERTY } = require('../common/constants')
 const { check } = require('express-validator')
 const { translateValidationMessage } = require('../common/translate-validation-message')
 
+const validateConfirmationCodeIsCorrect = () => (confirmationCode, { req }) => req.body.confirmationCode === req.session[CONFIRMATION_CODE_SESSION_PROPERTY]
+
 const validate = [
-  check('confirmationCode')
-    .not().isEmpty()
+  check('confirmationCode').custom(validateConfirmationCodeIsCorrect())
```
gethyn1: The outer function can be removed here and this line would become:
```
check('confirmationCode').custom(validateConfirmationCodeIsCorrect)
```

src/web/routes/application/enter-code/validate.js
```diff
@@ -1,9 +1,11 @@
+const { CONFIRMATION_CODE_SESSION_PROPERTY } = require('../common/constants')
 const { check } = require('express-validator')
 const { translateValidationMessage } = require('../common/translate-validation-message')
 
+const validateConfirmationCodeIsCorrect = () => (confirmationCode, { req }) => req.body.confirmationCode === req.session[CONFIRMATION_CODE_SESSION_PROPERTY]
```
gethyn1: Could `isCorrect` be removed here, it seems to be implied by `validate`. So function would be called `validateConfirmationCode()`

dwybourn: Updated to `validateConfirmationCode()`

src/web/routes/application/enter-code/validate.js
dwybourn: Removed outer function.

PR #322, opened 2019-07-10
HTBHF-1375 Content update for enter code.

dwybourn:


src/web/views/enter-code.njk
```diff
@@ -33,6 +33,6 @@
     {% endcall %}
 
     {# TODO DW HTBHF-1712 update link to send new confirmationCode  #}
-    <p class="govuk-body-m"><a href="#">{{ requestNewCode }}</a> {{ ifItDoesNotArrive }}</p>
+    <p class="govuk-body-m"><a href="#">{{ requestANewCode }}</a> {{ ifItDoesNotArrive }}</p>
```
gethyn1: I think the translation key `requestANewCode` could have remained as `requestNewCode` then this update is not required. I personally don't think the key names need to reflect the text exactly.

PR #323, opened 2019-07-10
HTBHF-1701 Updated 'Request a new code' link to redirect to send code…

dwybourn:
Request a new code link now takes the user back to send code page.

### DIFF
--- a/src/test/common/page/page.js
+++ b/src/test/common/page/page.js
@@ -124,15 +124,6 @@ class Page {
     }
   }
 
-  async findByLinkText (linkText) {
-    try {
-      return await this.driver.findElement(webdriver.By.linkText(linkText))
-    } catch (error) {
-      console.log(error)
-      throw new Error(error)
-    }
-  }
-
   async findH1 () {
     return this.findByCSS('h1')
   }
--- a/src/test/common/steps/enter-code-steps.js
+++ b/src/test/common/steps/enter-code-steps.js
@@ -27,5 +27,5 @@ Then(/^I am informed that I must enter in the code that was sent to me$/, async 
 Then(/^The request a new code link points to the send code page$/, async function () {
   const requestNewCodeLink = await pages.enterCode.getRequestNewCodeLink()
   const requestNewCodeLinkHref = await requestNewCodeLink.getAttribute('href')
-  expect(requestNewCodeLinkHref).to.be.equal(`${pages.url}/send-code`)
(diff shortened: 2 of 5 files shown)


src/test/common/steps/enter-code-steps.js
```diff
@@ -23,3 +23,9 @@ Then(/^I am informed that I must enter in the code that was sent to me$/, async
     pages.enterCode.getConfirmationCodeLinkErrorCss(),
     'Enter the 6 digit code we sent you')
 })
+
+Then(/^The request a new code link points to the send code page$/, async function () {
+  const requestNewCodeLink = await pages.enterCode.getRequestNewCodeLink()
+  const requestNewCodeLinkHref = await requestNewCodeLink.getAttribute('href')
+  expect(requestNewCodeLinkHref).to.be.equal(`${pages.url}/send-code`)
```
mikeSmith08: You should be getting the `/send-code` link from the SendCode page object.

src/web/views/enter-code.njk
```diff
@@ -33,6 +33,6 @@
     {% endcall %}
 
     {# TODO DW HTBHF-1712 update link to send new confirmationCode  #}
-    <p class="govuk-body-m"><a href="#">{{ requestANewCode }}</a> {{ ifItDoesNotArrive }}</p>
+    <p class="govuk-body-m"><a href="/send-code">{{ requestANewCode }}</a> {{ ifItDoesNotArrive }}</p>
```
mikeSmith08: Can we add an id to this link so we can use `findById` in `Page`? We then don't need the new method `findByLinkText`.

src/test/common/page/page.js
```diff
@@ -124,6 +124,15 @@ class Page {
     }
   }
 
+  async findByLinkText (linkText) {
```
mikeSmith08: See my comment below about adding an id to the link, we then don't need this method and can use `findById` which will be less fragile should the text on the link change.

src/test/common/steps/enter-code-steps.js
dwybourn: Updated to use `pages.sendCode.getPath()`

src/web/views/enter-code.njk
dwybourn: Added id.

src/test/common/page/page.js
dwybourn: Removed method.

PR #327, opened 2019-07-11
HTBHF-1712 Resend code if a user changes their contact details used t…

dwybourn:
After entering a confirmation code, the user now has to confirm a new code if they change the contact details used to receive a code.
i.e. changing your phone number will cause a new code to be send, only if the original code was texted.

Acceptance tests to come in next PR.

### DIFF
--- a/src/web/routes/application/common/confirmation-code.js
+++ b/src/web/routes/application/common/confirmation-code.js
@@ -1,7 +1,7 @@
-// reset confirmation code and reset state to in progress so that the user is not taken straight back to check details.
 const { CONFIRMATION_CODE_ENTERED_SESSION_PROPERTY } = require('./constants')
 const { stateMachine, states } = require('../common/state-machine')
 
+// reset confirmation code and reset state to in progress so that the user is not taken straight back to check details.
 const handleConfirmationCodeReset = req => {
   req.session[CONFIRMATION_CODE_ENTERED_SESSION_PROPERTY] = false
   req.session.claim.confirmationCode = null
--- a/src/web/routes/application/common/confirmation-code.js
+++ b/src/web/routes/application/common/confirmation-code.js
@@ -1,0 +1,16 @@
+// reset confirmation code and reset state to in progress so that the user is not taken straight back to check details.
+const { CONFIRMATION_CODE_ENTERED_SESSION_PROPERTY } = require('./constants')
+const { stateMachine, states } = require('../common/state-machine')
+
+const handleConfirmationCodeReset = req => {
+  req.session[CONFIRMATION_CODE_ENTERED_SESSION_PROPERTY] = false
+  req.session.claim.confirmationCode = null
+  stateMachine.setState(states.IN_PROGRESS, req)
+}
+
+const getConfirmationCodeChannel = req => req.session.claim.channelForCode
+
+module.exports = {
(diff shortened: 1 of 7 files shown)


src/web/routes/application/common/confirmation-code.js
```diff
@@ -0,0 +1,16 @@
+// reset confirmation code and reset state to in progress so that the user is not taken straight back to check details.
```
YetAnotherMatt: Shouldn't this comment be against the handleConfirmationCodeReset method?

dwybourn: Fixed.

YetAnotherMatt: it hasn't moved yet...

src/web/routes/application/common/confirmation-code.js
```diff
@@ -0,0 +1,16 @@
...
+const { CONFIRMATION_CODE_ENTERED_SESSION_PROPERTY } = require('./constants')
+const { stateMachine, states } = require('../common/state-machine')
+
+const handleConfirmationCodeReset = req => {
+  req.session[CONFIRMATION_CODE_ENTERED_SESSION_PROPERTY] = false
```
gethyn1: Not for this PR, but I think some of the higher level constants files could be split out into distinct modules (e.g. paths, session keys etc..).

src/web/routes/application/email-address/email-address.test.js
```diff
@@ -1,6 +1,14 @@
 const test = require('tape')
+const sinon = require('sinon')
+const proxyquire = require('proxyquire')
 const { contentSummary } = require('./email-address')
 
+const { TEXT, EMAIL } = require('../common/constants')
+
+const handleConfirmationCodeReset = sinon.spy()
+
+const { behaviourForPost } = proxyquire('./email-address', { '../common/confirmation-code': { handleConfirmationCodeReset } })
```
gethyn1: I know PRs are not really about formatting so not required, but I would maybe reformat this to multiple lines as its quite a lot of inline `{ ... }` so a bit tricky to determine the nesting.

src/web/views/enter-code.njk
```diff
@@ -34,5 +34,5 @@
 
     {# TODO DW HTBHF-1712 update link to send new confirmationCode  #}
     <p class="govuk-body-m"><a href="/send-code" id="request-new-code">{{ requestANewCode }}</a> {{ ifItDoesNotArrive }}</p>
```
gethyn1: Is there a way of putting the link in the translation? This implementation assumes that all languages will use the same order of words.

src/web/views/enter-code.njk
```diff
@@ -34,5 +34,5 @@
...
-s
+
```
gethyn1: I must have missed this in a previous PR - is there a way of putting the link in the translation? This implementation assumes that all languages will use the same order of words.

src/web/routes/application/common/confirmation-code.js
dwybourn: Not sure what happened there, fixed now.

src/web/views/enter-code.njk
dwybourn: Will do in a separate PR on it's own.

PR #328, opened 2019-07-12
HTBHF-1680 Add functions for removing all child entries from the session

mikeSmith08:
Add functions for removing all child entries from the session once the remove button has been clicked. Process is: 
- Get all the entries from the session that start with 'childDob-'
- Remove all entries which are for the given index ('childDob-x')
- Rebuild the object with the remaining children entries re-indexed plus everything else (non-'childDob-x' entries)
- This will be handed back to be added back into the session.

NOTE - this PR does not plumb it in yet and change what's in the session, this will be done in a subsequent PR.

### DIFF
--- a/src/web/routes/application/children-dob/decrement-keys.js
+++ b/src/web/routes/application/children-dob/decrement-keys.js
@@ -1,9 +1,9 @@
 const { isNil } = require('ramda')
 
-const INDEX_FROM_KEY_REG_EXP = /.*-(\d+).*/
+const KEY_INDEX_REG_EXP = /.*-(\d+).*/
 
 const getIndexFromKey = (key) => {
-  const groups = INDEX_FROM_KEY_REG_EXP.exec(key)
+  const groups = KEY_INDEX_REG_EXP.exec(key)
 
   if (isNil(groups)) {
     throw new Error(`Could not find index for key ${key}`)
--- a/src/web/routes/application/children-dob/decrement-keys.js
+++ b/src/web/routes/application/children-dob/decrement-keys.js
@@ -1,0 +1,34 @@
+const { isNil } = require('ramda')
+
+const INDEX_FROM_KEY_REG_EXP = /.*-(\d+).*/
+
+const getIndexFromKey = (key) => {
+  const groups = INDEX_FROM_KEY_REG_EXP.exec(key)
+
+  if (isNil(groups)) {
+    throw new Error(`Could not find index for key ${key}`)
+  }
+
+  return parseInt(groups[1], 10)
+}
+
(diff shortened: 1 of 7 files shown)


src/web/routes/application/children-dob/decrement-keys.js
```diff
@@ -0,0 +1,34 @@
+const { isNil } = require('ramda')
+
+const INDEX_FROM_KEY_REG_EXP = /.*-(\d+).*/
```
YetAnotherMatt: Minor, but maybe call this  'KEY_INDEX_REG_EXP'?

src/web/routes/application/children-dob/decrement-keys.test.js
```diff
@@ -0,0 +1,34 @@
...
+  t.equal(isKeyAfterRemoved('childDob-5', 1), true)
+
+  t.throws(() => isKeyAfterRemoved('childDob-day', 1), /Could not find index for key childDob-day/)
+  t.throws(() => isKeyAfterRemoved('inputCount', 1), /Could not find index for key inputCount/)
+  t.end()
+})
+
+test('decrementKey()', (t) => {
+  t.equal(decrementKey('childDob-2-day'), 'childDob-1-day')
+  t.equal(decrementKey('childDob-11-day'), 'childDob-10-day')
```
YetAnotherMatt: Might be worth adding a test for 'childDob-10-day' to confirm that the change from 2 digits to 1 is handled correctly

src/web/routes/application/children-dob/selectors.js
```diff
@@ -0,0 +1,14 @@
+const { pickBy } = require('ramda')
+const { isNotChildEntry, isChildEntry, keyDoesNotContainIndex } = require('./predicates')
+
+const getNonChildEntries = pickBy(isNotChildEntry)
+
+const getChildEntries = pickBy(isChildEntry)
+
+const getEntriesWithoutIndex = (entries, index) => pickBy(keyDoesNotContainIndex(index), entries)
```
gethyn1: This could be `omitKeysWithIndex` as this is similar to the function of Ramda's `omit` .

src/web/routes/application/children-dob/predicates.test.js
```diff
@@ -0,0 +1,22 @@
+const test = require('tape')
+const { keyDoesNotContainIndex, isNotChildEntry, isChildEntry } = require('./predicates')
+
+test('keyDoesNotContainIndex()', (t) => {
+  t.equal(keyDoesNotContainIndex(1)('value', 'childDob-1-day'), false)
```
YetAnotherMatt: can we test: `t.equal(keyDoesNotContainIndex(1)('value', 'childDob-11-day'), false)`?

PR #329, opened 2019-07-12
HTBHF-1801 Improved presentation of change links in send-code page

YetAnotherMatt:


src/web/server/locales/en/common.json
```diff
@@ -91,7 +91,7 @@
     "heading": "We’re sending you a code",
     "hint": "Select how you would like to receive it",
     "explanation": "We need this to confirm your identity and to prevent fraud.",
-    "change": "change"
+    "change": "Change"
```
YetAnotherMatt: I decided to keep these separate from the Change text on the check your answers page for flexibility. Let me know if you think they should be the same key.

PR #330, opened 2019-07-12
HTBHF-1705 Add acceptance tests for the Remove Child button.

mikeSmith08:
Currently have the main remove child test @ignore until functionality is complete. Testing of the presence and absence of remove buttons has been added to existing add child tests.

### DIFF
--- a/src/test/common/page/page.js
+++ b/src/test/common/page/page.js
@@ -85,11 +85,8 @@ class Page {
     }
   }
 
-  async findAllByClassName (className, waitFirst = true) {
+  async findAllByClassName (className) {
     try {
-      if (waitFirst) {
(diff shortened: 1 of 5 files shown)


src/test/common/page/page.js
```diff
@@ -85,9 +85,11 @@ class Page {
     }
   }
 
-  async findAllByClassName (className) {
+  async findAllByClassName (className, waitFirst = true) {
```
gethyn1: Are there any situations where we actually want this to wait for elements to exist? I guess so as its already in the code, but just asking the question.

mikeSmith08: Actually, there aren't, I've removed the waiting from both the findAll methods as it isn't needed and actually gets in the way.

PR #331, opened 2019-07-12
HTBHF-1735 Updated translation link for enter code page.

dwybourn:
All of the html for the request new code link has been put into a translation file as other languages may have different ordering of words and the two parts of the sentence translated separately may not make sense when concatenated.

### DIFF
--- a/src/web/routes/application/enter-nino/enter-nino.js
+++ b/src/web/routes/application/enter-nino/enter-nino.js
@@ -6,7 +6,7 @@ const pageContent = ({ translate }) => ({
   heading: translate('enterNino.heading'),
   ninoLabel: translate('enterNino.ninoLabel'),
   hint: translate('enterNino.hint'),
-  explanation: translate('enterNino.explanation', { ninoLink: 'https://www.gov.uk/lost-national-insurance-number' }),
+  explanation: translate('enterNino.explanation'),
   buttonText: translate('buttons:continue')
 })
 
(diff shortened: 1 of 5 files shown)


src/web/routes/application/enter-nino/enter-nino.js
```diff
@@ -6,7 +6,7 @@ const pageContent = ({ translate }) => ({
   heading: translate('enterNino.heading'),
   ninoLabel: translate('enterNino.ninoLabel'),
   hint: translate('enterNino.hint'),
-  explanation: translate('enterNino.explanation', { ninoLink: 'https://www.gov.uk/lost-national-insurance-number' }),
```
dwybourn: This link is no longer used.

src/web/server/locales/cy/common.json
```diff
@@ -99,11 +99,10 @@
     "weJustSentASixDigitCodeTo": "Lorem ipsum 6 dolor sit amet",
     "itMayTakeAFewMinutesToArrive": "Do eiusmod tempor incididunt.",
     "codeLabel": "Oced",
-    "requestANewCode": "Lorem ipsum dolor",
-    "ifItDoesNotArrive": "sunt in culpa qui officia."
+    "requestNewCodeLink": "<a href=\"/send-code\" id=\"{{requestNewCodeLinkId}}\">Consectetur adipiscing elit</a> consectetur adipiscing elit"
```
gethyn1: Why does the ID need to be a variable?

dwybourn: So it's not repeated across the translation files.

PR #334, opened 2019-07-15
HTBHF-1680 enable remove action for children's DOB

gethyn1:
- Handle remove action when remove action is present on POST
- Refactor `behaviourForPost` into three distinct middleware functions (`setChildrenInSessionForPost`, `handleAddAction`, `handleRemoveAction`)
- Update unit tests to reflect this refactor

The main functional changes exist in `handleRemoveAction.js`. The remaining updates are mainly refactoring code to separate files.

### DIFF
--- a/src/web/routes/application/children-dob/handle-add-action.js
+++ b/src/web/routes/application/children-dob/handle-add-action.js
@@ -1,6 +1,6 @@
-const { PATH } = require('./constants')
+const { PATH, ADD_CHILD_KEY } = require('./constants')
 
-const addActionRequested = body => body.hasOwnProperty('add')
+const addActionRequested = body => body.hasOwnProperty(ADD_CHILD_KEY)
 
 const handleAddAction = (req, res, next) => {
   if (!addActionRequested(req.body)) {
--- a/src/web/routes/application/children-dob/handle-add-action.js
+++ b/src/web/routes/application/children-dob/handle-add-action.js
@@ -1,0 +1,16 @@
+const { PATH } = require('./constants')
+
+const addActionRequested = body => body.hasOwnProperty('add')
+
+const handleAddAction = (req, res, next) => {
+  if (!addActionRequested(req.body)) {
+    return next()
+  }
+
+  req.session.children.inputCount = req.session.children.childCount + 1
+  return res.redirect(PATH)
+}
+
+module.exports = {
+  handleAddAction
+}
--- a/src/web/routes/application/children-dob/handle-remove-action.js
+++ b/src/web/routes/application/children-dob/handle-remove-action.js
@@ -1,0 +1,28 @@
+const { removeChildByIndex } = require('./remove-child-by-index')
+const { PATH } = require('./constants')
+
+const removeActionRequested = body => body.hasOwnProperty('remove')
+
+const decrementInputCount = childCount => childCount === 1 ? 1 : childCount - 1
+
+const handleRemoveAction = (req, res, next) => {
+  if (!removeActionRequested(req.body)) {
+    return next()
+  }
+
+  const { childCount } = req.session.children
+  const index = req.body.remove
+
+  if (index <= 0) {
+    throw new Error(`Unable to remove index ${index}`)
+  }
+
+  req.session.children = removeChildByIndex(req.session.children, index)
+  req.session.children.inputCount = decrementInputCount(childCount)
+  req.session.children.childCount = childCount - 1
+  return res.redirect(PATH)
+}
+
+module.exports = {
+  handleRemoveAction
+}
--- a/src/web/routes/application/children-dob/handle-remove-action.js
+++ b/src/web/routes/application/children-dob/handle-remove-action.js
@@ -1,7 +1,7 @@
 const { removeChildByIndex } = require('./remove-child-by-index')
-const { PATH } = require('./constants')
+const { PATH, REMOVE_CHILD_INDEX_KEY } = require('./constants')
 
-const removeActionRequested = body => body.hasOwnProperty('remove')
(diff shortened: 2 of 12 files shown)


src/web/routes/application/children-dob/set-children-in-session.js
```diff
@@ -0,0 +1,35 @@
+const { pickBy } = require('ramda')
+const { countKeysContainingString } = require('./count-keys')
+const { DAY_FIELD_SUFFIX } = require('./constants')
+
+const extractChildrenEntries = (val, key) => key.startsWith('child')
```
mikeSmith08: Should this be `keyStartsWithChild` rather than `extraChildrenEntries` as it doesn't actually extract anything, its just the predicate?

src/web/routes/application/children-dob/handle-add-action.js
```diff
@@ -0,0 +1,16 @@
+const { PATH } = require('./constants')
+
+const addActionRequested = body => body.hasOwnProperty('add')
+
+const handleAddAction = (req, res, next) => {
+  if (!addActionRequested(req.body)) {
```
dwybourn: Can we change this so the if statement is a positive rather than a negative. i.e. `if (addActionRequested(req.body))`. Reads much nicer that way.

src/web/routes/application/children-dob/set-children-in-session.js
gethyn1: Yes very good @mikeSmith08. I noticed this previously but didn't update as it was existing code. Now might be a good time to change this.

src/web/routes/application/children-dob/handle-remove-action.js
```diff
@@ -0,0 +1,28 @@
...
+
+const decrementInputCount = childCount => childCount === 1 ? 1 : childCount - 1
+
+const handleRemoveAction = (req, res, next) => {
+  if (!removeActionRequested(req.body)) {
+    return next()
+  }
+
+  const { childCount } = req.session.children
+  const index = req.body.remove
```
dwybourn: Could we update `remove` to something like `indexToRemove` or something similar. `remove` on it's own sounds like a flag to me.

src/web/routes/application/children-dob/handle-add-action.js
YetAnotherMatt: Presumably that would look like this:
```
const handleAddAction = (req, res, next) => {
  if (addActionRequested(req.body)) {
    req.session.children.inputCount = req.session.children.childCount + 1
    return res.redirect(PATH)
  }
  return next()
}
```

mikeSmith08: I normally prefer it the way it currently is, have the alternate flow to return/error quickly and then have the main body of the code to be for the positive case that the method was written for rather than be be completely within an if block.

gethyn1: If we followed the suggested approach in the remove middleware we would end up with nested if blocks, and it makes sense that they follow the same format. Anyone feel like voting on this??

src/web/routes/application/children-dob/handle-remove-action.js
gethyn1: Possibly, I guess it makes sense in the context of the middleware (although I think the variable name `index` makes it obvious what the purpose is), but I think it makes the HTML more confusing:
```
<button value="1" name="indexToRemove" type="submit" id="remove-child-1">Remove</button>
```
Does this mean the add button would also need updating?
```
<input type="submit" value="Add another child" name="add" id="add-another-child">
```

src/web/routes/application/children-dob/handle-add-action.js
gethyn1: This also means that all logic for the intended function would need to be inside an `if` block .. on second thoughts I'm not sure I agree with the suggestion, although I get the point about how it reads.

dwybourn: For the remove middleware I'd leave it as is which is why I didn't leave a comment there.

I'd argue readability is more important, not to mention the `if(!condition) { next() }` format isn't consistent with existing middleware.

dwybourn: I accept this is very much a preference matter so don't really mind if you prefer the way it is.

mikeSmith08: If we're voting, I'd leave it as it is, I don't agree with Dave here I'm afraid.

src/web/routes/application/children-dob/handle-remove-action.js
gethyn1: Maybe:
```
<name="removeChildIndex" value="1" />
<name="addChild" value="Add another child" />
```
?

mikeSmith08: I kind of agree with Dave, but I'm tempted to leave it as it is because by making the JS better, we make the HTML worse, so its a flip of a coin. So I'd go with leaving it as it requires no further change.

src/web/routes/application/children-dob/set-children-in-session.js
gethyn1: Turns out we already had a function for this `isChildEntry`

src/web/routes/application/children-dob/handle-add-action.js
gethyn1: @dwybourn if you're not too fussed then I'm going to leave this as is.

src/web/routes/application/children-dob/handle-remove-action.js
gethyn1: I've updated to use `removeChildIndex` - I've also updated the HTML to use `button` instead of `input` to be consistent with GOV.UK button component.

PR #335, opened 2019-07-15
HTBHF-1686 added scss to control the position of the remove child button, and add a section-break between children

YetAnotherMatt:


src/web/views/macros/htbhf-child-dob-input.njk
```diff
@@ -10,50 +10,55 @@
 
 {% macro htbhfChildDobInput(params) %}
 
-  {% call govukFieldset({
-      legend: {
-        text: params.text.legend,
-        classes: "govuk-fieldset__legend--m"
-      }
-   }) %}
+  <div class="c-htbhf-child-dob-input">
```
YetAnotherMatt: I added the div and the hr, all else is just a change to the indentation

src/web/assets/sass/settings/_media-queries.scss
```diff
@@ -0,0 +1,10 @@
+/**
+ * Breakpoint definitions
+ * Defensively define breakpoints for different devices
+ */
+
+$govuk-breakpoints: (
```
gethyn1: Looks like too much indentation here

src/web/views/macros/htbhf-child-dob-input.njk
```diff
@@ -10,50 +10,55 @@
...
+          value: params.index,
+          name: "remove",
+          attributes: {
+            id: "remove-child-" + params.index
+          }
+        }) }}
+      </div>
+    {% endif %}
+  </div>
+  <hr class="govuk-section-break govuk-section-break--xl govuk-section-break--visible">
```
gethyn1: I think possibly the `hr` should be in the parent template. In this file it means every time the component is used it will have a line at the bottom. My thoughts are the line is required due to stacked / multiple child inputs which is a feature of the parent component. This component would be more reusable if it didn't assume this structure / style itself.

Currently, if a design decision was made to remove the `hr` from the last instance of the component in the list it would be tricky to deal with - however it should be trivial if it was in the parent component (which has knowledge of the render loop).

YetAnotherMatt: done

src/web/assets/sass/settings/_media-queries.scss
YetAnotherMatt: Indentation reduced

src/web/views/macros/htbhf-child-dob-input.njk
YetAnotherMatt: Now just the div - the hr has moved

PR #337, opened 2019-07-16
HTBHF-1705 Refactor existing page functions that get or enter values …

mikeSmith08:
…into input fields to use common page functions.

### DIFF
--- a/src/test/common/page/enter-children-dob.js
+++ b/src/test/common/page/enter-children-dob.js
@@ -9,6 +9,12 @@ const PAGE_TITLES = {
   cy: 'GOV.UK - Urna condimentum mattis?'
 }
 
+const getDayInputIdForIndex = (index) => `childDob-${index}-day`
+const getMonthInputIdForIndex = (index) => `childDob-${index}-month`
+const getYearInputIdForIndex = (index) => `childDob-${index}-year`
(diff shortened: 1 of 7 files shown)


src/test/common/page/enter-children-dob.js
```diff
@@ -9,6 +9,12 @@ const PAGE_TITLES = {
   cy: 'GOV.UK - Urna condimentum mattis?'
 }
 
+const getChildDobDayInputIdForIndex = (index) => `childDob-${index}-day`
+const getChildDobMonthInputIdForIndex = (index) => `childDob-${index}-month`
+const getChildDobYearInputIdForIndex = (index) => `childDob-${index}-year`
+const getChildDobNameInputIdForIndex = (index) => `childDob-name-${index}`
+const getRemoveChildButtonIdForIndex = (index) => `remove-child-${index}`
```
YetAnotherMatt: Minor suggestion (that shouldn't stop this being merged, and is a matter of opinion): As these are only used within this file the names could be a bit shorter, e.g. `getChildDobDayInputIdForIndex` -> `getDayInputIdForIndex`

mikeSmith08: Yep, good suggestion, done.

PR #338, opened 2019-07-16
HTBHF-1704 create partition utility and implement in removeChildByIndex

gethyn1:
Created `partitionObj` function which creates two objects from a single object based on a predicate.

`partitionObj` has been implemented in `removeChildByIndex` which means that most of the child DOB selectors can be removed. It also makes the function slightly more efficient by splitting the object in a single pass.


src/web/routes/application/children-dob/remove-child-by-index.js
```diff
@@ -11,12 +14,12 @@ const reIndexChildren = (children, index) => compose(fromPairs, map(handleDecrem
  *  - Rebuild the object with the remaining children entries re-indexed plus everything else (non-'childDob-x' entries)
  */
 const removeChildByIndex = (children, index) => {
-  const childEntries = getChildEntries(children)
-  const filteredChildren = omitKeysWithIndex(childEntries, parseInt(index, 10))
+  const partitioned = partitionObj(isChildEntry, children)
```
dwybourn: Was the reason for this change efficiency, or are there plans to use `partitionObj` in other places? 
Right now I'd argue that the way it was before reads nicer, `childEntries` vs `partitioned[0]`

gethyn1: Updated so `partitionedObj` now returns a single object in the format:
```
{
  matches: { ... },
  others: { ... }
}
```

PR #339, opened 2019-07-16
HTBHF-1704 Fix finding the phone number field by id after merge with master.

mikeSmith08:


src/test/common/page/phone-number.js
```diff
@@ -27,6 +27,11 @@ class PhoneNumber extends SubmittablePage {
     return 'phone number'
   }
 
+  // TODO MS: HTBHF-1836 Refactor this and clearPhoneNumber into InputField Class
+  async getPhoneNumberField () {
+    return this.findById('phone-number')
+  }
```
mikeSmith08: This function was deleted on the story branch but used in a merge to master. This just shows we always need to rebase a story branch before merging any PRs into it.

PR #346, opened 2019-07-17
HTBHF-1844 Added new form description macro. Updated content for phon…

dwybourn:
- Added new macro for form description. (Black text underneath H1 text)
- Phone number page is now using a form description instead of a hint (grey text).
- Updated content to match prototype.

### DIFF
--- a/src/web/server/locales/en/common.json
+++ b/src/web/server/locales/en/common.json
@@ -68,7 +68,7 @@
   "phoneNumber": {
     "title": "What’s your mobile telephone number?",
     "heading": "What’s your mobile telephone number?",
-    "hint": "This must be the same number you use for any benefits you claim.",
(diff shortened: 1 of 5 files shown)


src/web/server/locales/en/common.json
```diff
@@ -68,7 +68,7 @@
   "phoneNumber": {
     "title": "What’s your mobile telephone number?",
     "heading": "What’s your mobile telephone number?",
-    "hint": "This must be the same number you use for any benefits you claim.",
+    "formDescription": "If you claim any benefits you must use the same address registered with the benefits agency (DWP or HMRC).",
```
mikeSmith08: Is this correct, should we be using the word `address` when this is on the phone number page?

dwybourn: Well spotted. Chatted with Sam and 'address' has been updated with 'mobile number'. This will be added in the prototype shortly.

PR #355, opened 2019-07-18
Group summary lists by list name

gethyn1:
Update the structure of the data consumed by the “check details” page to allow rendering multiple summary lists:

- Group the data passed to the view by list
- Refactor the summary list into a macro
- Refactor the page content structure to allow for easy translation of the summary list headings

### DIFF
--- a/src/web/routes/application/check/get-row-data.test.js
+++ b/src/web/routes/application/check/get-row-data.test.js
@@ -1,5 +1,5 @@
 const test = require('tape')
-const { getRowData, getFlattenedRowData } = require('./get-row-data')
+const { getRowData, getFlattenedRowData, groupRowData } = require('./get-row-data')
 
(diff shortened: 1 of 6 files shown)


src/web/routes/application/check/get-row-data.test.js
```diff
@@ -67,3 +68,83 @@ test('getFlattenedRowData returns flattened row data with step with empty conten
...
+    ]
+  }
+
+  const result = groupRowData(rowData)
+
+  t.deepEqual(result, expected, 'returns an object with row data grouped by list')
+  t.end()
+})
+
+test('getGroupedRowData returns grouped row data', (t) => {
```
mikeSmith08: Feels like this is two tests in one, testing that the grouping by list is performed, and then that if no list key is provided then the default list key will be used. Maybe worth splitting out into two tests?

PR #356, opened 2019-07-19
HTBHF-1467 Update successful confirmation page to match prototype and update ATs accordingly.

mikeSmith08:


src/web/server/locales/en/common.json
```diff
@@ -116,7 +116,7 @@
   "confirm": {
     "newClaimTitle": "Application complete",
     "updatedClaimTitle": "Application updated",
-    "subTitle": "You are entitled to\n£{{totalVoucherValue}} per week"
+    "subTitle": "You are entitled to\n£{{totalVoucherValue}} a week. Your first payment\nwill be £{{totalVoucherValueForFourWeeks}}."
```
dwybourn: The prototype is using 'You're' instead of 'You are'

dwybourn: The monetary values need to be bold.

src/web/routes/application/confirm/confirm.js
```diff
@@ -31,8 +31,11 @@ const getConfirmPage = (req, res, next) => {
 
     return res.render('confirm', {
       title: getTitle(req),
-      subTitle: req.t('confirm.subTitle', { totalVoucherValue: toPounds(totalVoucherValueInPence) }),
-      totalVoucherValueForFourWeeks: toPounds(totalVoucherValueInPence * 4)
+      subTitle: req.t('confirm.subTitle',
+        {
+          totalVoucherValue: toPounds(totalVoucherValueInPence),
+          totalVoucherValueForFourWeeks: toPounds(totalVoucherValueInPence * 4)
```
YetAnotherMatt: This might not be the correct value :(
If they have applied just before a child turns 4 this won't be correct. It's not been introduced as part of this commit though, and the communication around such a  scenario would be really tricky, so let's leave it for now...

src/web/server/locales/en/common.json
mikeSmith08: Fixed.

src/web/routes/application/confirm/confirm.js
mikeSmith08: Yes, that sounds like an edge case which we probably wouldn't want to complicate this page with. Might be worthwhile raising with Karen and/or Sam and see whether this needs to be documented somewhere like the Business Rules document?

PR #357, opened 2019-07-19
Check details for “Do you have any children?”

gethyn1:
- Add “do you have any children?” to check details page
- Invalidate review if user changes answer for “do you have any children?” from “no” to “yes”

### DIFF
--- a/src/web/routes/application/common/state-machine/state-machine.js
+++ b/src/web/routes/application/common/state-machine/state-machine.js
@@ -21,9 +21,7 @@ const getStepForPath = (path, steps) => steps.find(step => step.path === path)
  * Next path is navigable if the next step does not exist, if it doesn't define an isNavigable function, or that function returns true.
  * E.g. if the current step is the end of the 'in-progress' journey, the next path will be /check. There is no step matching /check, so we assume it is navigable.
  */
-function isNextPathNavigable (nextStep, req) {
-  return !nextStep || !nextStep.isNavigable || nextStep.isNavigable(req.session)
-}
+const isNextPathNavigable = (nextStep, req) => !nextStep || !nextStep.isNavigable || nextStep.isNavigable(req.session)
 
 /**
  * Ask the current step for the next path. Test whether the step matching that path is navigable. If not, ask that step for the next path; repeat.
@@ -32,18 +30,18 @@ function isNextPathNavigable (nextStep, req) {
  * @param steps the steps of the apply journey
  * @returns the path of the next step
  */
-const getNextPath = (path, req, steps) => {
+const getNextNavigablePath = (path, req, steps) => {
   const thisStep = getStepForPath(path, steps)
   const nextPath = getNextForStep(req, thisStep)
   const nextStep = getStepForPath(nextPath, steps)
   if (isNextPathNavigable(nextStep, req)) {
     return nextPath
   }
-  return getNextPath(nextPath, req, steps)
+  return getNextNavigablePath(nextPath, req, steps)
 }
 
 const getNext = (req, steps) => {
-  return getNextPath(req.path, req, steps)
+  return getNextNavigablePath(req.path, req, steps)
 }
 
 const isPathAllowed = (sequence, allowed, path) =>
--- a/src/web/routes/application/common/state-machine/state-machine.js
+++ b/src/web/routes/application/common/state-machine/state-machine.js
@@ -17,9 +17,33 @@ const actions = {
 
 const getStepForPath = (path, steps) => steps.find(step => step.path === path)
 
+/**
+ * Next path is navigable if the next step does not exist, if it doesn't define an isNavigable function, or that function returns true.
+ * E.g. if the current step is the end of the 'in-progress' journey, the next path will be /check. There is no step matching /check, so we assume it is navigable.
+ */
+function isNextPathNavigable (nextStep, req) {
+  return !nextStep || !nextStep.isNavigable || nextStep.isNavigable(req.session)
+}
+
+/**
+ * Ask the current step for the next path. Test whether the step matching that path is navigable. If not, ask that step for the next path; repeat.
+ * @param path the path of the current step
+ * @param req the current request
+ * @param steps the steps of the apply journey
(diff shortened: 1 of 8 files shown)


src/test/acceptance/features/do-you-have-children.feature
```diff
@@ -3,28 +3,37 @@ Feature: Do you have children who are 3 years old or younger?
...
     And I click continue
     Then I am shown the enter your childrens dates of birth page
 
   Scenario: Children’s dates of birth is not navigable via the back button if I’ve said I have no children
+    Given I have entered my details up to the do you have children three or younger page
     When I have said No to the do you have children three or younger question
     And I am shown the are you pregnant page
     Then The back link points to the Do you have children who are three years old or younger page
+
+  Scenario: I can change my answer to yes for 'Do you have children?' from the check details page
```
mikeSmith08: As discussed, let's move this out into its own feature file, something like `change-children.feature` then we can reinstate the `Background` for this feature file.

src/web/routes/application/common/state-machine/state-machine.test.js
```diff
@@ -21,6 +25,13 @@ test(`Dispatching ${GET_NEXT_PATH} should return should return next property of
   t.end()
 })
 
+test(`Dispatching ${GET_NEXT_PATH} should return should return next navigable path when state of ${states.IN_PROGRESS} defined in session and next step is not navigable`, async (t) => {
```
mikeSmith08: Minor, you have `should return should return` here.

src/web/routes/application/common/state-machine/state-machine.js
```diff
@@ -17,9 +17,33 @@ const actions = {
 
 const getStepForPath = (path, steps) => steps.find(step => step.path === path)
 
+/**
+ * Next path is navigable if the next step does not exist, if it doesn't define an isNavigable function, or that function returns true.
+ * E.g. if the current step is the end of the 'in-progress' journey, the next path will be /check. There is no step matching /check, so we assume it is navigable.
+ */
+function isNextPathNavigable (nextStep, req) {
```
gethyn1: Should we use this format to be consistent:
```
const isNextPathNavigable = (nextStep, req) => !nextStep || !nextStep.isNavigable || nextStep.isNavigable(req.session)
```
```

File: src/web/routes/application/common/state-machine/state-machine.js [hunk 4]
```diff
@@ -17,9 +17,33 @@ const actions = {
...
+}
+
+/**
+ * Ask the current step for the next path. Test whether the step matching that path is navigable. If not, ask that step for the next path; repeat.
+ * @param path the path of the current step
+ * @param req the current request
+ * @param steps the steps of the apply journey
+ * @returns the path of the next step
+ */
+const getNextPath = (path, req, steps) => {
```
gethyn1: I think maybe this should be `getNextNavigablePath`, which would also help understanding why `getNext` is just wrapping another function.

src/web/routes/application/common/state-machine/state-machine.js
```diff
@@ -17,9 +17,33 @@ const actions = {
 
 const getStepForPath = (path, steps) => steps.find(step => step.path === path)
 
+/**
+ * Next path is navigable if the next step does not exist, if it doesn't define an isNavigable function, or that function returns true.
+ * E.g. if the current step is the end of the 'in-progress' journey, the next path will be /check. There is no step matching /check, so we assume it is navigable.
+ */
+function isNextPathNavigable (nextStep, req) {
+  return !nextStep || !nextStep.isNavigable || nextStep.isNavigable(req.session)
```
gethyn1: Maybe unnecessary but `isNil(nextStep) || isNil(nextStep.isNavigable)` is more explicit than `!nextStep ...`.

`nextStep.isNavigable(req.session)` is the only check where we are expecting a boolean.

src/web/routes/application/common/state-machine/state-machine.js
mikeSmith08: Yeah, I don't see the point in the `getNext` function if all its doing it calling `getNextPath`.

YetAnotherMatt: I've renamed `getNextPath` to `getNextNavigablePath`. There's good reason for `getNext` to continue its existence though. Otherwise this sort of configuration would be quite a bit more complex:
```
    getNextPath: getNext,
```

gethyn1: @YetAnotherMatt you could always inline that function:
```
getNextPath : (req, steps) => getNextNavigablePath(req.path, req, steps)
```

YetAnotherMatt: What's the verdict on whether I need to do so? (I pushed the latest changes before reading this final comment)

gethyn1: I think inlining is a good idea - it will still be a single line and I find the function wrapping an unnecessary abstraction.

src/web/routes/application/common/state-machine/state-machine.test.js
```diff
@@ -4,7 +4,11 @@ const { CHECK_URL, TERMS_AND_CONDITIONS_URL, CONFIRM_URL } = require('../../comm
 
 const { GET_NEXT_PATH, INVALIDATE_REVIEW } = actions
 
-const steps = [{ path: '/first', next: () => '/second' }, { path: '/second' }]
+const steps = [
+  { path: '/first', next: () => '/second' },
+  { path: '/second', next: () => '/third' },
+  { path: '/third', isNavigable: () => false, next: () => '/fourth' }
```
dwybourn: Being picky here, but can we have the fourth entry on it's own line.

mikeSmith08: That's not a fourth entry, its the `next` for the third entry.

YetAnotherMatt: There is no fourth entry (deliberately). The '/fourth' path is actually the next property of the third step.

PR #359, opened 2019-07-22
Display children’s dates of birth on “check details” page

gethyn1:
- Add ignored acceptance test for changing answer from “yes” to “no”, the code to make this pass will be in a subsequent PR (this can be removed from this PR if it is confusing)
- Update HTML structure and acceptance tests for “check details” page to allow the children’s dates of birth summary list to be tested independently (as it’s HTML structure will differ from the main summary list)
- Wrap GOV.UK summary list macro to create a custom summary list for children’s dates of birth

### DIFF
--- a/src/web/views/macros/htbhf-children-summary-list.njk
+++ b/src/web/views/macros/htbhf-children-summary-list.njk
@@ -1,0 +1,47 @@
+{% from "summary-list/macro.njk" import govukSummaryList %}
+
+{#
+  htbhfChildDobInput() Wraps the GOV.UK summary list component to provide a single action for multiple
+  summary list rows
+#}
+{% macro htbhfChildrenSummaryList(params) %}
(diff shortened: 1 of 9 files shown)


src/web/views/macros/htbhf-children-summary-list.njk
```diff
@@ -0,0 +1,47 @@
...
+{#
+  htbhfChildDobInput() Wraps the GOV.UK summary list component to provide a single action for multiple
+  summary list rows
+#}
+{% macro htbhfChildrenSummaryList(params) %}
+  {% set childRows = [] %}
+
+  {% for i in range(0, params.children.childCount) %}
+    {% set index = i + 1 %}
+    {% set actions = [] %}
```
YetAnotherMatt: Don't think we're using `actions` any more?

PR #361, opened 2019-07-22
HTBHF-1793 Added autofocus to the last child name when there's more t…

dwybourn:
Added autofocus for child's name on child dob page. Only added for last entry and if there's more than one input.

Went for an approach different to that mentioned in the jira story. Happy to change it if we feel another way is better.


src/web/views/children-dob.njk
```diff
@@ -25,7 +26,8 @@
         legend: aboutYourChild,
         dateLegend: dateOfBirth,
         removeButton: removeChild
-      }
+      },
+      autofocus: i == children.inputCount - 1 and children.inputCount > 1
```
gethyn1: This should use strict equality operator `===`

gethyn1: Does this introduce it's own a11y issues? When a user with more than one child chooses to change details or use the back button, they will always be focused on the last input.

dwybourn: Just thought of that and addressing that now.

dwybourn:
> Happy for this to go in as is, but I'd like to do a bit more in another PR. Can we do something like adding a `req.session.children.autofocus=n` property in `handleAddAction`, and unset that property in `setChildrenInSessionForPost`? As `behaviourForPost` is defined as:
> 
> ```
> 
> const behaviourForPost = [
>   setChildrenInSessionForPost,
>   handleAddAction,
>   handleRemoveAction
> ]
> ```
> 
> I think that means that the property would be unset on every post, and reset if the action is to add a child.

@YetAnotherMatt Added as part of this PR.

src/web/views/children-dob.njk
dwybourn: Updated to ===

src/web/views/macros/htbhf-child-dob-input.njk
```diff
@@ -17,6 +17,7 @@
         }
      }) %}
 
+
```
gethyn1: Rogue new line

dwybourn: Appears to be a git error.

PR #362, opened 2019-07-22
Display children’s dates of birth on “check details page”

gethyn1:
- Wrap GOV.UK summary list macro to create a custom summary list for children’s dates of birth
- Display children’s dates of birth on “check details page”
- Update HTML structure and acceptance tests for “check details” page to allow the children’s dates of birth summary list to be tested independently (as it’s HTML structure will differ from the main summary list)

Adding acceptance tests for this update is quite involved so leaving for a separate (subsequent) PR.

### DIFF
--- a/src/web/routes/application/check/get.js
+++ b/src/web/routes/application/check/get.js
@@ -12,7 +12,8 @@ const pageContent = ({ translate }) => ({
   summaryListHeadings: {
     aboutYou: translate('check.aboutYou'),
     aboutYourChildren: translate('check.aboutYourChildren')
-  }
+  },
+  childrensDobHiddenText: translate('childrenDob.summaryKey')
 })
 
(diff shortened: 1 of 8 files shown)


src/web/routes/application/check/get.js
```diff
@@ -32,7 +33,7 @@ const getCheck = (steps) => (req, res) => {
   req.session.nextAllowedStep = stateMachine.dispatch(actions.GET_NEXT_PATH, req, steps)
 
   res.render('check', {
-    claim: req.session.claim,
+    children: req.session.children,
```
mikeSmith08: Where does the rest of the claim come from for the other fields on the check page if this has changed to `children` rather than `claim`?

gethyn1: Correct - `claim` was never used. All other data comes from `checkRowData`

PR #367, opened 2019-07-24
Format date for children’s dates of birth summary

gethyn1:
- Move building of rows for children’s dates of birth summary list out of the view (this means we can reuse date formatters without creating a new Nunjucks filter. It also follows how the rows for the other summary lists are generated)
- Format children’s dates of birth in summary list
- Import correct date formatter in acceptance tests

### DIFF
--- a/src/web/routes/application/check/get-childrens-dates-of-birth-rows.js
+++ b/src/web/routes/application/check/get-childrens-dates-of-birth-rows.js
@@ -1,5 +1,4 @@
 const { formatDateForDisplay } = require('../common/formatters')
-const { NAME_KEY, DATE_OF_BIRTH_KEY } = require('./constants')
 
 const buildRow = (keyText, valueText) => ({
   key: {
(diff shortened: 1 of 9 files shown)


src/web/routes/application/check/get-childrens-dates-of-birth-rows.js
```diff
@@ -0,0 +1,36 @@
...
+  )
+
+const getChildrensDatesOfBirthRows = (children) => {
+  const rows = []
+
+  for (let i = 0; i < children.childCount; i++) {
+    const index = i + 1
+    const nameRow = buildRow(NAME_KEY, children[`childDobName-${index}`])
+    const dateOfBirthRow = buildRow(DATE_OF_BIRTH_KEY, getFormattedDateForChild(children, index))
+    rows.push(nameRow)
```
dwybourn: This can just be rows.push(nameRow, dateOfBirthRow)

PR #370, opened 2019-07-24
Remove children from session

gethyn1:
Remove children from session when changing answer for “Do you have children?” from “Yes” to “No”.

Checking children's existence on session is now implemented using `isNil` which allows for checking both `null` and `undefined`. An alternative would be to remove `children` using the `delete` operator.

### DIFF
--- a/src/web/routes/application/children-dob/set-children-in-session.js
+++ b/src/web/routes/application/children-dob/set-children-in-session.js
@@ -1,10 +1,10 @@
-const { pickBy } = require('ramda')
+const { pickBy, isNil } = require('ramda')
 const { countKeysContainingString } = require('./count-keys')
 const { DAY_FIELD_SUFFIX } = require('./constants')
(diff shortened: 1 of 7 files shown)


src/web/routes/application/children-dob/set-children-in-session.js
```diff
@@ -1,10 +1,10 @@
-const { pickBy } = require('ramda')
+const { pickBy, isNil } = require('ramda')
 const { countKeysContainingString } = require('./count-keys')
 const { DAY_FIELD_SUFFIX } = require('./constants')
 const { isChildEntry } = require('./predicates')
 
 const setChildrenInSessionForGet = (req) => {
-  if (!req.session.hasOwnProperty('children')) {
+  if (isNil(req.session.children)) {
```
YetAnotherMatt: we need top treat children in the same way (use isNil) in create-request-body when posting to the claimant service

PR #373, opened 2019-07-25
HTBHF-1670 rename do you have children files

gethyn1:
Step 1 of renaming “Do you have children three or under?”. The end result after all PRs will be:

Path: /do-you-have-children


src/test/common/page/do-you-have-children.js
```diff
@@ -0,0 +1,38 @@
+'use strict'
```
gethyn1: This is just a file rename. Not sure why Github has not seen this.

PR #374, opened 2019-07-25
HTBHF-1670 rename do you have children features

gethyn1:
Step 2 of renaming “Do you have children three or under?”. The end result after all PRs will be:

Path: /do-you-have-children


src/test/common/steps/navigation.js
```diff
@@ -21,7 +21,7 @@ const ENTER_NAME_PAGE = 'enter name'
 const DO_YOU_LIVE_IN_SCOTLAND_PAGE = 'do you live in Scotland'
 const ENTER_NINO_PAGE = 'enter national insurance'
 const ENTER_DOB_PAGE = 'enter date of birth'
-const DO_YOU_HAVE_CHILDREN_THREE_OR_YOUNGER_PAGE = 'do you have children three or younger'
+const DO_YOU_HAVE_CHILDREN_THREE_OR_YOUNGER_PAGE = 'do you have children under four years old'
```
mikeSmith08: Should this constant be renamed too, or is that the next PR?

gethyn1: Its in the next PR

PR #375, opened 2019-07-25
HTBHF-1670 rename references to do you have children

gethyn1:
Step 3 of renaming “Do you have children three or under?”. The end result after all PRs will be:

Path: /do-you-have-children

### DIFF
--- a/src/test/common/steps/common-steps.js
+++ b/src/test/common/steps/common-steps.js
@@ -117,19 +117,19 @@ async function enterEmailAddressAndSubmit (emailAddress = EMAIL_ADDRESS) {
   }
 }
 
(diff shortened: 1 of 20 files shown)


src/test/common/steps/common-steps.js
```diff
@@ -117,19 +117,19 @@ async function enterEmailAddressAndSubmit (emailAddress = EMAIL_ADDRESS) {
   }
 }
 
-async function selectNoOnChildrenThreeOrYoungerPage () {
+async function selectNoOnChildrenPage () {
```
mikeSmith08: This should probably be `selectNoOnDoYouHaveChildrenPage`. Similarly with the yes equivalent below.

PR #379, opened 2019-07-29
HTBHF-1909 rename template caching variable

gethyn1:


manifest.yml
```diff
@@ -8,7 +8,6 @@ applications:
     CLAIMANT_SERVICE_URL: http://htbhf-claimant-service((space-suffix)).apps.internal:8080
     NODE_ENV: production
     APP_VERSION: ((app-version))
-    CACHE_VIEW_TEMPLATES: true
```
dwybourn: This should be renamed to NO_CACHE_VIEW_TEMPLATES instead of being removed.

gethyn1: But it will be set to false

dwybourn: I see, no point of setting it then.

PR #381, opened 2019-07-30
Create basic guidance pages structure

gethyn1:
- Created base Nunjucks template for guidance pages to extend from
- Registered route for first guidance page. This route will eventually change but is indicative of how all the guidance page routes will be registered
- Pass navigation meta data to the view

Future PRs to story branch will:
- Register other guidance page routes
- Deal with translations
- Display page content
- Display contents navigation

### DIFF
--- a/src/web/routes/guidance.js
+++ b/src/web/routes/guidance.js
@@ -1,0 +1,71 @@
+const { compose, equals, prop } = require('ramda')
+const { handleRequestForPath } = require('./application/middleware')
+
+const pages = [
+  {
+    title: 'How it works',
+    path: '/how-it-works'
+  },
+  {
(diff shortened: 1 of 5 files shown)


src/web/routes/guidance.js
```diff
@@ -0,0 +1,71 @@
...
+    title: 'Apply for Healthy Start',
+    path: '/apply-for-healthy-start'
+  },
+  {
+    title: 'Report a change',
+    path: '/report-a-change'
+  }
+]
+
+const isPathForPage = (path) => compose(equals(path), prop('path'))
```
mikeSmith08: I don't think this method name is quite right, how about something like `hasMatchingPath`?

PR #394, opened 2019-08-01
HTBHF-1915 Added validation of county field.

dwybourn:
Added validation for acceptance tests for county address field.

### DIFF
--- a/src/test/acceptance/features/manual-address.feature
+++ b/src/test/acceptance/features/manual-address.feature
@@ -20,6 +20,10 @@ Feature: Address
     When I do not enter the second line of an address
     Then I am shown the phone number page
 
+  Scenario: Enter a valid address without an optional county line
+    When I do not enter the 'county' of an address
+    Then I am shown the phone number page
+
   Scenario Outline: Enter an address with an invalid postcode
     When I enter an address with postcode <postcode>
     Then I am informed that the postcode is in the wrong format
@@ -42,9 +46,6 @@ Feature: Address
     When I do not enter the 'town or city' of an address
     Then I am informed that I need to enter an address on the 'town or city' field
 
-  Scenario: Do not enter in the county
-    When I do not enter the 'county' of an address
-    Then I am informed that I need to enter an address on the 'county' field
 
   Scenario: Enter in an address where the first line is too long
     When I enter in an address where the first line is too long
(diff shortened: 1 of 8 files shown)


src/test/acceptance/features/manual-address.feature
```diff
@@ -42,6 +42,10 @@ Feature: Address
     When I do not enter the 'town or city' of an address
     Then I am informed that I need to enter an address on the 'town or city' field
 
+  Scenario: Do not enter in the county
```
YetAnotherMatt: I'd rather not make County a required field - we could default to a sensible value ('UK'? - do this in the claimant service, not here) if it's not provided.

dwybourn: It's required for allpay, and will be printed on the letters. Maybe one for the UX team?

gethyn1: What @YetAnotherMatt said!!

YetAnotherMatt: I've spoken to the UX team, their preference is for it not to be there at all, and if it does have to be then for it to not be mandatory. If it's not present in the OS places response we won't be asking them to add it manually, so we'll have to handle an empty value anyway. Let's have the County field for now, not mandatory, and default to 'UK' in the claimant service if it's not provided. We can revisit whether the field is necessary once we've got OS places and postcodes.io integrated.

dwybourn: Ok, will make optional

mikeSmith08: Can we make sure that we have a test that makes sure an application is successful with no county then please?

src/test/common/steps/manual-address-steps.js
```diff
@@ -61,6 +70,11 @@ Then(/^I am informed that I need to enter an address on the 'town or city' field
   await assertTownOrCityErrorFieldAndLink('Tell us your address')
 })
 
+Then(/^I am informed that I need to enter an address on the 'county' field$/, async function () {
```
mikeSmith08: You can remove this method when it becomes optional.

src/test/acceptance/features/manual-address.feature
dwybourn: Field is now optional. Complete application test will come in next PR due to number of files it involves changing.

src/test/acceptance/features/manual-address.feature
```diff
@@ -42,6 +46,7 @@ Feature: Address
     When I do not enter the 'town or city' of an address
     Then I am informed that I need to enter an address on the 'town or city' field
 
+
```
gethyn1: Rogue new line

dwybourn: Will remove in next PR.

PR #398, opened 2019-08-02
HTBHF-1982 Added failure event logging if we cannot call notify

YetAnotherMatt:


src/web/routes/application/send-code/notify.js
```diff
@@ -6,26 +6,40 @@ const { logger } = require('../../../logger')
...
   const reference = uuid()
   const notifyOptions = { personalisation: { confirmationCode: code }, reference: reference }
 
-  logConfirmationCodeEvent(channel, reference)
+  const logData = { eventType: SEND_2FA_CODE, timestamp: new Date(), codeChannel: channel, notifyReference: reference }
+  logger.info(JSON.stringify(logData))
 
   if (channel === EMAIL) {
     notifyClient.sendEmail(EMAIL_TEMPLATE_ID, claim.emailAddress, notifyOptions)
+      .catch(e => { logFailureEvent(logData, 'Error sending email', e) })
```
gethyn1: The curly braces aren't required here:
```
.catch(e => logFailureEvent(logData, 'Error sending email', e))
```

PR #399, opened 2019-08-02
HTBHF-1964 Add acceptance tests for guidance pages.

mikeSmith08:
- The page object for guidance pages is a bit different because it encapsulates all the guidance pages so has its own navigation step and function.
 - The tests themselves assert that the headers are correct, that there a links for all pages (table of contents) apart from the page itself, plus there are previous and next links if applicable. There are no assertions for actual content.

This was my best stab at getting the feature file correct, please feel free to suggest alternatives, I found it quite hard to get them right so they might not be perfect yet.

### DIFF
--- a/src/test/common/page/guidance.js
+++ b/src/test/common/page/guidance.js
@@ -1,5 +1,5 @@
 'use strict'
-
+const { head, last } = require('ramda')
 const Page = require('./page')
 
 const { PAGES, getPageMetadata } = require('../../../web/routes/guidance')
@@ -20,7 +20,7 @@ class Guidance extends Page {
   }
 
   getPageTitle (pageName) {
-    return 'GOV.UK - ' + pageName
+    return `GOV.UK - ${pageName}`
   }
 
   getPageWithTitle (pages, pageName) {
@@ -28,11 +28,11 @@ class Guidance extends Page {
   }
 
   hasPreviousLink (pageName) {
-    return pageName !== 'How it works'
+    return pageName !== head(PAGES).title
   }
 
   hasNextLink (pageName) {
-    return pageName !== 'Report a change'
+    return pageName !== last(PAGES).title
   }
 
   getNextLink (pageName) {
@@ -59,9 +59,9 @@ class Guidance extends Page {
 
   // Find links in the contents section, which should be for all the pages except the current page
   async getContentsLinks (pageName) {
-    const allPageExceptCurrent = PAGES.filter(page => page.title !== pageName)
-    const pagePath = allPageExceptCurrent.map(async (page) => this.findLinkForHref(this.getPageMetadataForPage(page.title).activePath))
-    return Promise.all(pagePath)
+    const allPagesExceptCurrent = PAGES.filter(page => page.title !== pageName)
+    const pagePaths = allPagesExceptCurrent.map(async (page) => this.findLinkForHref(this.getPageMetadataForPage(page.title).activePath))
+    return Promise.all(pagePaths)
   }
 
   async findLinkForHref (href) {
--- a/src/test/common/page/guidance.js
+++ b/src/test/common/page/guidance.js
@@ -1,0 +1,73 @@
+'use strict'
+
+const Page = require('./page')
+
+const { PAGES, getPageMetadata } = require('../../../web/routes/guidance')
+
+/**
+ * Page object for all the guidance pages. This may extend Page, but as it represent multiple static page, it must
+ * implement its own methods for navigation, but can use all the find methods from Page when required.
+ */
+class Guidance extends Page {
+  async openGuidancePage (baseURL, pageName, lang = 'en') {
+    try {
+      await this.openPage(`${baseURL}${this.getPageWithTitle(PAGES, pageName).path}`, lang)
+      return super.waitForPageWithTitle(this.getPageTitle(pageName))
+    } catch (error) {
+      console.error(`Error caught trying to open guidance page at: ${baseURL}`, error)
+      throw new Error(error)
+    }
+  }
+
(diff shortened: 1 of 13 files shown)


src/test/common/page/guidance.js
```diff
@@ -0,0 +1,73 @@
...
+      await this.openPage(`${baseURL}${this.getPageWithTitle(PAGES, pageName).path}`, lang)
+      return super.waitForPageWithTitle(this.getPageTitle(pageName))
+    } catch (error) {
+      console.error(`Error caught trying to open guidance page at: ${baseURL}`, error)
+      throw new Error(error)
+    }
+  }
+
+  getPageTitle (pageName) {
+    return 'GOV.UK - ' + pageName
```
gethyn1: Could be:
```
return `GOV.UK - ${pageName}`
```

as you've used this format on above lines

src/test/common/page/guidance.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  getPageWithTitle (pages, pageName) {
+    return pages.find(page => page.title === pageName)
+  }
+
+  hasPreviousLink (pageName) {
+    return pageName !== 'How it works'
```
gethyn1: Could these methods check against first / last page titles in `PAGES` array? rather than be static strings?

src/test/common/page/guidance.js
mikeSmith08: Done.

src/test/common/page/guidance.js
mikeSmith08: Good point, done.

src/test/common/page/guidance.js
```diff
@@ -0,0 +1,73 @@
...
+
+  async findPreviousLinkForPage (pageName) {
+    const previousLinkHref = this.getPreviousLink(pageName).path
+    return this.findLinkForHref(previousLinkHref)
+  }
+
+  // Find links in the contents section, which should be for all the pages except the current page
+  async getContentsLinks (pageName) {
+    const allPageExceptCurrent = PAGES.filter(page => page.title !== pageName)
+    const pagePath = allPageExceptCurrent.map(async (page) => this.findLinkForHref(this.getPageMetadataForPage(page.title).activePath))
```
gethyn1: Should this be changed `pagePath` -> `pagePaths`?

src/test/common/steps/guidance-steps.js
```diff
@@ -0,0 +1,46 @@
...
+})
+
+async function assertGuidancePageHeadersCorrect (pageName) {
+  const h1Text = await pages.cookies.getH1Text()
+  expect(h1Text.toString().trim()).to.be.equal('Get money off milk, food and vitamins (Healthy Start)',
+    'expected guidance page H1 text to be correct')
+  const h2Text = await pages.cookies.getH2Text()
+  let expectedH2PageName = pageName
+  if (pageName === 'What you get') {
+    // Page title from pages doesn't match what's in the title of the What you get page
```
gethyn1: Does this mean the `PAGES` object is incorrect for “What you get”? Can we just update `PAGES`?

src/web/routes/guidance/index.js
```diff
@@ -1,3 +1,5 @@
 module.exports = {
-  ...require('./guidance')
+  ...require('./guidance'),
+  ...require('./pages'),
+  ...require('./get-page-meta-data')
```
gethyn1: This exports every function from `./get-page-meta-data` - it might be better to only expose the functions that you require:

```
const { getPageMeta } = require('./get-page-meta-data')

module.exports {
  ...require('./guidance'),
  ...require('./pages'),
  getPageMeta
}
```

src/test/common/steps/guidance-steps.js
mikeSmith08: No, I tried this but tests fail sadly, I'm not sure whether it's worth trying to fix those tests or leave this like this, what do you think?

src/test/common/page/guidance.js
mikeSmith08: Yep, good spot, done.

src/web/routes/guidance/index.js
mikeSmith08: Fixed.

src/test/common/steps/guidance-steps.js
gethyn1: I would try and fix as I don't think its great to have this sort of check. Maybe the page objects in `PAGES` need updating to:
```
{
  title: 'What you’ll get',
  path: '/what-you-get',
  label: 'What you get'
}
```

mikeSmith08: Ok, it turned out to be easier than I thought to fix this, its sorted now and this extra check has been removed.

src/test/common/steps/guidance-steps.js
```diff
@@ -0,0 +1,41 @@
...
+
+Then(/^all the (.*) guidance page content is present$/, async function (pageName) {
+  await assertGuidancePageHeadersCorrect(pageName)
+  await assertGuidancePageContentsCorrect(pageName)
+  await assertGuidancePageNavigationLinksCorrect(pageName)
+})
+
+async function assertGuidancePageHeadersCorrect (pageName) {
+  const h1Text = await pages.cookies.getH1Text()
+  expect(h1Text.toString().trim()).to.be.equal('Get money off milk, food and vitamins (Healthy Start)',
```
YetAnotherMatt: is the H1 text the same on every guidance page?

mikeSmith08: Yes.

PR #400, opened 2019-08-02
Guidance page translations

gethyn1:
- Rename previous and next parameters to differentiate between pages and labels
- Add translations for guidance pages
- Remove `title` from page meta as it’s not used in the template - this removes `getPageForPath` function and tests as no longer required

### DIFF
--- a/src/web/routes/guidance/get-page-meta-data.js
+++ b/src/web/routes/guidance/get-page-meta-data.js
@@ -1,7 +1,5 @@
 const { compose, equals, prop } = require('ramda')
 
-const getPageForPath = (pages, path) => pages.find(hasMatchingPath(path))
-
 const hasMatchingPath = (path) => compose(equals(path), prop('path'))
 
 const getPreviousPage = (pages, index) => pages[index - 1]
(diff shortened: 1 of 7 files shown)


src/web/server/locales/cy/common.json
```diff
@@ -154,9 +154,14 @@
...
   },
+  "guidance": {
+    "title": "Ex monetaria mooeo, vitus et grub (Healthy Start)",
+    "navigationHeading": "Contentus"
+  },
   "yes": "Diam",
   "no": "Sed",
   "back": "Kcab",
   "text": "Occaecat",
-  "email": "Lobortis"
```
dwybourn: Why was email removed? Guessing this wasn't intentional?

src/web/routes/guidance/get-page-meta-data.js
```diff
@@ -10,18 +8,15 @@ const getNextPage = (pages, index) => pages[index + 1]
 
 const getPageMetadata = (pages, path) => {
   const pageIndexForPath = pages.findIndex(hasMatchingPath(path))
-  const page = getPageForPath(pages, path)
 
   return {
     activePath: path,
-    previous: getPreviousPage(pages, pageIndexForPath),
-    next: getNextPage(pages, pageIndexForPath),
-    title: page.title
```
YetAnotherMatt: purely out of curiosity, why has title been removed?

src/web/server/locales/cy/common.json
gethyn1: Errrr .. no, that was a mistake. Have updated.

PR #404, opened 2019-08-05
HTBHF-1972 Updated to latest version of selenium and enabled config switch for headless mode

YetAnotherMatt:
The latest version of selenium seems to work with headless chrome, so the switch I've added here isn't strictly necessary. Seems nicer than having to edit code to change between modes though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,18 +8,16 @@ cache:
   directories:
     - bin/
 
+dist: bionic
+
 services:
 - redis-server
 
-dist: trusty
-
 addons:
   chrome: stable
-  firefox: latest
 
 before_install:
 - google-chrome-stable --headless --disable-gpu --remote-debugging-port=9222 http://localhost &
-- firefox -headless &
 - npm install -g standard tape nyc
 
 deploy:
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -8,18 +8,16 @@ cache:
   directories:
     - bin/
 
+dist: bionic
```
mikeSmith08: What does this do?

src/test/common/config.js
```diff
@@ -7,11 +7,13 @@ const SESSION_DETAILS_BASE_URL = process.env.SESSION_DETAILS_BASE_URL || `http:/
 const SESSION_DETAILS_PATH = '/session-details/confirmation-code'
 
 module.exports.PORT = process.env.PORT
-module.exports.TEST_BROWSER = process.env.TEST_BROWSER || 'chrome'
+module.exports.TEST_BROWSER = 'chrome'
```
YetAnotherMatt: I'll revert this once all branches have inherited this and we can remove the env setting in travis

mikeSmith08: Is there any point in keeping this in? Same with the value in `sample.env` too.

.travis.yml
YetAnotherMatt: Changes the version of ubuntu used by travis. Bionic is the most recent LTS. It was needed for xvfb, but we don't need xvfb now...

PR #405, opened 2019-08-05
HTBHF-1966 Add accessibility tests for guidance pages.

mikeSmith08:


src/test/a11y/results.js
```diff
@@ -11,7 +11,7 @@ const handleTestResults = results => {
     resultsWithIssues.forEach(console.error)
     process.exit(1)
   } else {
-    console.log(`No accessibility issues found, (${results.length} results in total)`)
+    console.log(`No accessibility issues found (${results.length} results in total)`)
```
YetAnotherMatt: If we're going to remove the comma, how about `No accessibility issues found in (${results.length} results in total)`?

mikeSmith08: The number of results is in brackets which is why I removed the comma, if the brackets were removed then I'd keep the comma. Your suggestion also works without the brackets, but I'll leave it like that with the brackets in place.

PR #410, opened 2019-08-06
Remove overview page

gethyn1:
Remove “Overview” page and replace with “Guidance” page as first page of the application process:

- Remove overview page from app
- Register apply guidance page on root URL
- Remove overview page from acceptance tests
- Update acceptance tests to use apply guidance page as first page


src/test/common/page/guidance.js
```diff
@@ -4,14 +4,23 @@ const Page = require('./page')
...
 class Guidance extends Page {
+  constructor (driver) {
+    super(driver)
+    this.sessionId = null
+  }
+
   async openGuidancePage (baseURL, pageName, lang = 'en') {
     try {
       await this.openPage(`${baseURL}${this.getPageWithTitle(PAGES, pageName).path}`, lang)
+      this.sessionId = await this.getCurrentSessionId()
```
gethyn1: Guidance pages to not use base page `open` method so session ID needs to be stored at this point

src/test/common/page/page.js
```diff
@@ -32,7 +32,6 @@ class Page {
   async open (baseURL, lang = 'en') {
     try {
       await this.openDirect(baseURL, lang)
-      this.sessionId = await this.getCurrentSessionId()
```
mikeSmith08: You can also remove the line `this.sessionId = null` from the constructor in this file too.

src/test/smoke/features/smoke-test.feature
```diff
@@ -3,7 +3,7 @@ Feature: Smoke test
   As the continuous delivery pipeline
   I want know that a new release has deployed correctly and is functional
 
-  Scenario: The overview page is available and starts the application process
-    Given I navigate to the Apply for Healthy Start overview page
+  Scenario: The guidance page is available and starts the application process
```
mikeSmith08: I think this should start `The apply guidance page is available...`

src/test/common/page/page.js
gethyn1: Updated

src/test/smoke/features/smoke-test.feature
gethyn1: Updated

PR #422, opened 2019-08-08
Feature/handle path request on complete

gethyn1:
Updated how incoming requests are handled when the state of the apply journey is `COMPLETE` to be inclusive instead of exclusive, and pass through requests if no redirect required:

- Assert if path is valid by comparing with the paths in `steps` object (instead of assuming a redirect should happen)
- Redirect on `COMPLETE` state only if requested path exists in `steps`
- Move `handle-path-request.js` to separate directory
- Refactor `handle-path-request.js` by moving utility functions into separate files with the aim of improving legibility of the pertinent function


src/web/routes/application/middleware/handle-path-request/handle-path-request.js
```diff
@@ -1,20 +1,21 @@
...
     res.clearCookie('lang')
-    const redirectPath = isGuidancePageUrl(req.path) ? req.path : config.environment.OVERVIEW_URL
-    return res.redirect(redirectPath)
+
+    if (isPathInSequence(req.path, pathsInSequence)) {
+      return res.redirect(config.environment.OVERVIEW_URL)
+    }
+
+    return next()
   }
```
gethyn1: Lines 14 -> 18 are the important parts of this PR:

- Remove redirect by default - call `next()` if path is allowed
- Assert if requested path is **in** sequence instead of asserting for paths **outside** of the sequence

src/web/routes/application/middleware/handle-path-request/handle-path-request.js
```diff
@@ -1,20 +1,21 @@
...
-
 const middleware = (config, pathsInSequence, step) => (req, res, next) => {
   // Destroy the session on navigating away from CONFIRM_URL
   if (stateMachine.getState(req) === states.COMPLETED && req.path !== CONFIRM_URL) {
     req.session.destroy()
     res.clearCookie('lang')
-    const redirectPath = isGuidancePageUrl(req.path) ? req.path : config.environment.OVERVIEW_URL
-    return res.redirect(redirectPath)
+
+    if (isPathInSequence(req.path, pathsInSequence)) {
```
mikeSmith08: This doesn't read quite right, I know that's what the function does but I think it needs to be more descriptive of the scenario its dealing with so might need to be something like `isPathInApplicationFlow`. Let's discuss when you get in.

gethyn1: Renamed to `isPathInApplicationFlow`

PR #430, opened 2019-08-13
HTBHF-898 Content update for privacy notice.

dwybourn:
Content updated to match latest design doc https://docs.google.com/document/d/1W44YqlBqTZmwdwNL8tNSURe7Vc2_Cc2mGEz2Y5qqCvw/edit. 
At time of writing the prototype has not been updated so use the doc as reference for reviewing.

### DIFF
--- a/src/web/views/locales/cy/privacy-notice.njk
+++ b/src/web/views/locales/cy/privacy-notice.njk
@@ -15,7 +15,7 @@
   Sed interdum maximus <a href="https://www.gov.uk/government/organisations/department-of-health-and-social-care/about/personal-information-charter" target="_blank">
     sed interdum maximus</a> sed interdum maximus.</p>
 
-<p class="govuk-body">Sed interdum maximus sed interdum maximus.</p>
+<p class="govuk-body">Sed <a class="govuk-link" href="/cookies">interdum maximus</a> sed interdum maximus.</p>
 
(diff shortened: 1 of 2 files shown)


src/web/views/locales/cy/privacy-notice.njk
```diff
@@ -177,7 +177,7 @@
 
 <div class="c-htbhf-inset">
   <p>
-    The Information Commissioner<br>
+    Information Commissioner's Office<br>
```
gethyn1: Does this need to use `’` instead of `'` (subtle character difference)

src/web/views/locales/en/privacy-notice.njk
```diff
@@ -176,7 +176,7 @@
 
 <div class="c-htbhf-inset">
   <p>
-    The Information Commissioner<br>
+    Information Commissioner's Office<br>
```
gethyn1: See previous comment about punctuation

src/web/views/locales/cy/privacy-notice.njk
dwybourn: Well spotted, updated.

src/web/views/locales/en/privacy-notice.njk
dwybourn: Well spotted, updated.

PR #440, opened 2019-08-15
HTBHF-2065 Added feature toggle ability to the navigation functionality of acceptance tests.

mikeSmith08:
- Completely restructure AT navigation steps so that the page load and action are now associated with each other so that we can easily toggle out a single page.
 - Split out the navigation functions from navigation steps and add to their own directory.
 - Move page step actions out into their own file.
 - Tidy and refactor the navigation functions for better readability and simplicity.


src/test/common/steps/complete-flow.js
```diff
@@ -16,13 +16,10 @@ Given('I am on the first page of the application', async function () {
 })
 
 Given('I am on the check details page having entered valid details for a pregnant woman', async function () {
-  await pages.startApplication()
```
mikeSmith08: All the steps removed in this file were superfluous which meant that extra page navigation being done for no reason. Similarly the page waits are now always done by the navigation functions.

src/web/routes/register-steps/register-steps.js
```diff
@@ -11,10 +11,10 @@ const { toBooleanStrict } = require('./to-boolean-strict')
  * - toggle must be a string
  * - config must be a boolean (if it exists)
  *
- * N.B. shouldRegisterStep coerces string representations of 'true' and 'false' to boolean
+ * N.B. isStepEnabled coerces string representations of 'true' and 'false' to boolean
  * as it is assumed the config values will be set as environment variables.
  */
-const shouldRegisterStep = config => step => {
+const isStepEnabled = config => step => {
```
mikeSmith08: Renamed this function so that it made sense in terms of being used both by the main application and the acceptance tests.

PR #442, opened 2019-08-15
HTBHF-1973 Updated content of the unsuccessful application page

YetAnotherMatt:


src/web/views/guidance/en/what-you-get.njk
```diff
@@ -55,7 +55,7 @@
 
 <h3 class="govuk-heading-m">Vitamin vouchers</h3>
 
-<p class="govuk-body">Vitamins are not part of the card service yet. You’ll be sent vouchers for those separately which you can exchange for vitamins.</p>
+<p class="govuk-body">Vitamins are not part of the card service yet. You’ll be sent vouchers for them separately if you want them. You can exchange them for vitamins.</p>
```
YetAnotherMatt: This is an unrelated change arising from the review of HTBHF-1910

src/web/views/locales/en/unsuccessful/error.html
```diff
@@ -0,0 +1,3 @@
+<p class="govuk-body">An error occurred whilst checking your eligibility</p>
+<p class="govuk-body"> <a href="apply">Go back and try again</a>.</p>
```
dwybourn: Should the links have the govuk-link class attached? Or, should we finally update the css so we don't have to keep doing this.

src/web/views/locales/en/unsuccessful/pending.html
```diff
@@ -0,0 +1,2 @@
+<p class="govuk-body">Your benefit status is pending</p>
+<p class="govuk-body">If you’re still having problems call 0300 330 7010 and select option 3.</p>
```
YetAnotherMatt: Pending content hasn't been provided yet - I just added the phone number to the bottom of all decisions.

src/web/views/locales/en/unsuccessful/error.html
mikeSmith08: You've missed the class from the <a> tag.

PR #443, opened 2019-08-15
HTBHF-2067 set features in top level JSON file

gethyn1:
- Set features in top level JSON file to allow other services to easily consume features
- Import features JSON into environment config


src/web/routes/register-steps/register-steps.js
```diff
@@ -37,7 +38,10 @@ const isStepEnabled = config => step => {
  *
  * Removes steps from array that are toggled off based on logic in isStepEnabled filter
  */
-const registerSteps = (config, steps) => steps.filter(isStepEnabled(config))
+const registerSteps = (config, steps) => {
```
YetAnotherMatt: `registerSteps` doesn't seem to reflect what this method is actually doing, though it's not something that has changed in this pr.

PR #449, opened 2019-08-19
HTBHF-2043 Changed /do-you-live-in-scotland to /scotland

YetAnotherMatt:
This will need a commensurate change in the performance (and new acceptance) tests


src/test/common/page/in-scotland.js
```diff
@@ -10,13 +10,13 @@ const PAGE_TITLES = {
 /**
  * Page object for I Live in Scotland page where the claimant is told they cannot apply if they live in Scotland.
```
mikeSmith08: Minor - would it be worthwhile removing the `I Live in Scotland` words from this comment, it then still makes perfect sense but without something that tracks the name of the object?

src/test/a11y/tests/in-scotland.js
```diff
@@ -13,5 +13,5 @@ const iLiveInScotland = [
 ]
 
 module.exports = {
-  iLiveInScotland
+  iLiveInScotland: inScotland
```
gethyn1: There is an issue with this export - it should be:
```
module.exports = {
  inScotland
}
```
This might give you some issues elsewhere.

src/test/a11y/tests/index.js
```diff
@@ -1,5 +1,5 @@
 const { apply } = require('./apply')
-const { iLiveInScotland } = require('./i-live-in-scotland')
+const { iLiveInScotland } = require('./in-scotland')
```
gethyn1: See above comment - should be:
```
const { inScotland } = require('./in-scotland')
```

src/web/routes/application/in-scotland/in-scotland.js
```diff
@@ -2,26 +2,26 @@ const { path } = require('ramda')
...
+  path: '/in-scotland',
+  template: 'in-scotland',
   pageContent,
   isNavigable,
   behaviourForGet
 }
 
 module.exports = {
-  iLiveInScotland
+  iLiveInScotland: inScotland
```
gethyn1: Similar import / export issue as previous comment

src/web/routes/application/in-scotland/index.js
```diff
@@ -0,0 +1,5 @@
+const { iLiveInScotland } = require('./in-scotland')
```
gethyn1: Needs updating to match updated export. Could do the following to avoid future updates:
```
module.exports = require('./in-scotland')
```

src/web/routes/application/scotland/scotland.js
```diff
@@ -2,32 +2,32 @@ const { validate } = require('./validate')
...
-  template: 'do-you-live-in-scotland',
+  template: 'scotland',
   pageContent,
   contentSummary,
   validate
 }
 
 module.exports = {
-  doYouLiveInScotland
+  scotland: scotland
```
gethyn1: Could use the same shorthand syntax (as before this update):
```
module.exports = {
  scotland
}
```

PR #453, opened 2019-08-19
HTBHF-2047 Updated postcode lookup to use feature toggle and added a1…

dwybourn:
- Updated postcode to use feature toggle.
- Added acceptance and a11y tests for postcode page.


src/test/common/steps/navigation/step-page-actions.js
```diff
@@ -46,6 +47,10 @@ const STEP_PAGE_ACTIONS = [
     page: (pages) => pages.enterNino,
     actions: async () => enterNinoAndSubmit()
   },
+  {
+    page: (pages) => pages.postcode,
+    actions: async () => enterPostcodeAndSubmit()
```
mikeSmith08: You'll need to add the toggle into this step too otherwise when the page is toggled off this acceptance test will fail.

`toggle: 'ADDRESS_LOOKUP_ENABLED'`

dwybourn: Added toggle.

PR #454, opened 2019-08-20
HTBHF-2043 renamed /children-dob to /child-date-of-birth

YetAnotherMatt:


src/web/routes/application/child-date-of-birth/child-date-of-birth.js
```diff
@@ -50,5 +50,5 @@ const childrenDob = {
 module.exports = {
   behaviourForGet,
   behaviourForPost,
-  childrenDob
+  childDateOfBirth: childDateOfBirth
```
gethyn1: This can use shorthand syntax (as previous):
```
{
  ...
  childDateOfBirth
}
```

YetAnotherMatt: Sorry - thought I'd caught all of those

PR #476, opened 2019-08-21
HTBHF-2043 changed /enter-name to /name

YetAnotherMatt:


cd_test_versions.properties
```diff
@@ -0,0 +1,3 @@
+# This file is `source`d by the cd pipeline when running tests
```
YetAnotherMatt: Well, it will be. This file currently has no effect.

test_versions.properties
```diff
@@ -0,0 +1,3 @@
+# This file is `source`d by the cd pipeline when running tests
```
YetAnotherMatt: Well, it will be. This file does nothing yet.

PR #480, opened 2019-08-22
HTBHF-1649 update national insurance number path

gethyn1:
Update path for ”What’s your National Insurance number?”

- Rename path: `/enter-nino` -> `/national-insurance-number`
- Add missing word `number` from `national insurance number` in feature files


src/test/common/page/national-insurance-number.js
```diff
@@ -9,9 +9,9 @@ const PAGE_TITLES = {
 }
 
 /**
- * Page object for EnterNino page where the national insurance number is entered.
+ * Page object for NationalInsuranceNumber page where the national insurance number is entered.
```
mikeSmith08: How about just: `Page object for the page where the national insurance number is entered.`

src/web/routes/application/national-insurance-number/national-insurance-number.js
```diff
@@ -0,0 +1,29 @@
...
+  heading: translate('nationalInsuranceNumber.heading'),
+  ninoLabel: translate('nationalInsuranceNumber.ninoLabel'),
+  hint: translate('nationalInsuranceNumber.hint'),
+  explanation: translate('nationalInsuranceNumber.explanation'),
+  buttonText: translate('buttons:continue')
+})
+
+const contentSummary = (req) => ({
+  key: req.t('nationalInsuranceNumber.summaryKey'),
+  value: req.session.claim.nino
```
mikeSmith08: This still being `nino` looks like the proverbial sore thumb here...

gethyn1: Yes - I discussed with team and decided to leave the claim property as `nino`, but I don't mind renaming this as well - although would this have implications on the claimant service?

PR #485, opened 2019-08-27
Query OS places API for postcode

gethyn1:
- Lookup address from postcode using OS places API
- Adapt response from OS places API
- Add results of OS places API request to session
- Update acceptance tests to 1) set a default postcode 2) always clear input before entering values

If input is not cleared before entering a new value then the postcode entered for `/postcode` would be populated on `/manual-address` which results in breaking acceptance tests.


src/web/routes/application/postcode/adapters.js
```diff
@@ -0,0 +1,26 @@
...
+
+const DELIVERY_POINT_ADDRESS_PROP = 'DPA'
+
+const ADDRESS_FIELDS = [
+  'ADDRESS',
+  'ORGANISATION_NAME',
+  'BUILDING_NUMBER',
+  'SUB_BUILDING_NAME',
+  'BUILDING_NAME',
+  'THOROUGHFARE_NAME',
```
YetAnotherMatt: We should probably include DEPENDENT_THOROUGHFARE_NAME as well - CH53QW

src/web/routes/application/postcode/postcode.js
```diff
@@ -1,3 +1,29 @@
+const request = require('request-promise')
+const { wrapError } = require('../common/formatters')
+const { transformOsPlacesApiResponse } = require('./adapters')
+
+const OS_PLACES_API_URL = 'https://api.ordnancesurvey.co.uk/places/v1/addresses/postcode'
```
mikeSmith08: Might we need to make this configurable, potentially add it as an environment variable?

src/web/routes/application/postcode/postcode.test.js
```diff
@@ -0,0 +1,76 @@
...
+    t.equal(next.called, true, 'calls next()')
+  } catch (error) {
+    t.fail(error)
+  }
+
+  next.resetHistory()
+  t.end()
+})
+
+test('behaviourForPost() handles succesful address lookup', async (t) => {
```
mikeSmith08: Test description isn't correct here.

src/web/routes/application/postcode/postcode.test.js
```diff
@@ -0,0 +1,76 @@
...
+  ]
+
+  request.returns(Promise.resolve(TEST_FIXTURES))
+
+  try {
+    await behaviourForPost(config)(req, res, next)
+
+    t.deepEqual(req.session.postcodeLookupResults, expectedAdresses, 'adds postcode lookup results to session')
+    t.equal(next.called, true, 'calls next()')
+  } catch (error) {
```
mikeSmith08: We don't normally wrap tests in a try/catch, wouldn't it automatically fail the test if it threw an error?

src/web/routes/application/postcode/postcode.js
dwybourn: If you add `OS_PLACES_URI: getVCAPServiceOSPlacesVariable('OS_PLACES_URI', process.env.OS_PLACES_URI)` to environments.js that should get it.

src/web/routes/application/postcode/postcode.js
```diff
@@ -1,3 +1,29 @@
...
+
+const behaviourForPost = (config) => async (req, res, next) => {
+  const { OS_PLACES_API_KEY } = config.environment
+  const { postcode } = req.body
+
+  try {
+    const addressLookupResults = await request({
+      uri: `${OS_PLACES_API_URL}?postcode=${postcode}&key=${OS_PLACES_API_KEY}`,
```
dwybourn: This will need updating as the uri doesn't include any paths. Currently it's set to `https://api.ordnancesurvey.co.uk`

src/web/routes/application/postcode/postcode.js
```diff
@@ -1,3 +1,29 @@
...
+const { transformOsPlacesApiResponse } = require('./adapters')
+
+const OS_PLACES_API_URL = 'https://api.ordnancesurvey.co.uk/places/v1/addresses/postcode'
+
+const behaviourForPost = (config) => async (req, res, next) => {
+  const { OS_PLACES_API_KEY } = config.environment
+  const { postcode } = req.body
+
+  try {
+    const addressLookupResults = await request({
```
YetAnotherMatt: Can we specify a timeout?

src/web/routes/application/postcode/postcode.js
gethyn1: I might leave adding the URL to config for a separate PR

src/web/routes/application/postcode/postcode.js
gethyn1: Yes I think so. What timeout do you think is appropriate?

src/web/routes/application/postcode/postcode.test.js
```diff
@@ -0,0 +1,76 @@
+const test = require('tape')
+const sinon = require('sinon')
+const proxyquire = require('proxyquire')
+const TEST_FIXTURES = require('./test-fixtures.json')
+
+const request = sinon.stub()
+const { behaviourForPost } = proxyquire('./postcode', { 'request-promise': request })
+const config = { environment: { OS_PLACES_API_KEY: '123' } }
+
+test('behaviourForPost() handles succesful address lookup', async (t) => {
```
YetAnotherMatt: Are we missing a 'handles lookup failure' test?

src/web/routes/application/postcode/postcode.test.js
YetAnotherMatt: This needs renaming

src/web/routes/application/postcode/postcode.js
YetAnotherMatt: good question! 30 seconds at the _very_ longest, but maybe something like 10 seconds?

gethyn1: I think maybe 5s? If OS places takes 10s to respond then page requests will back up as the API request blocks the response from our server (to the user).

src/web/routes/application/postcode/adapters.js
gethyn1: Added this one.

src/web/routes/application/postcode/postcode.js
gethyn1: Added this to config

src/web/routes/application/postcode/postcode.test.js
gethyn1: Good spot - updated

src/web/routes/application/postcode/postcode.test.js
gethyn1: It's not strictly necessary, but I've added a comment about why I've added this. I can remove the try / catch if people prefer.

src/web/routes/application/postcode/postcode.js
gethyn1: I've updated this with value from config

src/web/routes/application/postcode/postcode.js
gethyn1: Added a 5s timeout - can update to 10s if you think that is better.

YetAnotherMatt: 5 seconds is good.

PR #489, opened 2019-08-28
HTBHF-2166 Added acceptance tests for select address.

dwybourn:
- Tests are just for address not found. Selecting an address is a future subtask.


src/test/common/steps/address-lookup-steps.js
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable no-unused-expressions */
+const { When, Then } = require('cucumber')
+const { expect } = require('chai')
+
+const pages = require('./pages')
+const { enterPostcodeAndSubmit } = require('./common-steps')
+const { setupPostcodeLookupWithNoResults } = require('../wiremock')
+
+const POSTCODE_WITH_NO_RESULTS = 'AA1 1AA'
```
YetAnotherMatt: Suggest using BS1 !AA to line up with the os places stub

PR #493, opened 2019-08-29
Feature/render address select

dwybourn:
- Now displaying a dropdown list of addresses after performing an address lookup (if there are results).


src/test/common/steps/address-lookup-steps.js
```diff
@@ -26,3 +32,13 @@ Then(/^I am shown a button to enter my address manually$/, async function () {
   const buttonText = await pages.selectAddress.getSubmitButtonText()
   expect(buttonText).to.be.equal('Enter address manually')
 })
+
+Then(/^I am shown a list of addresses$/, async function () {
+  const addressOptions = await pages.selectAddress.getAddressOptions()
+  expect(addressOptions.length).to.be.equal(11)
```
mikeSmith08: I think it might be worthwhile checking that the addresses returned are relevant, so maybe check that each address shown has the correct postcode?

dwybourn: Done.

src/web/routes/application/select-address/select-address.js
```diff
@@ -7,11 +7,17 @@ const pageContent = ({ translate }) => ({
   enterManualAddress: translate('buttons:enterManualAddress'),
   explanation: translate('address.explanation'),
   change: translate('change'),
-  continue: translate('buttons:continue')
+  continue: translate('buttons:continue'),
+  selectAddressLabel: translate('address.selectAddressLabel')
+})
+
+const buildAddressOption = result => ({
```
YetAnotherMatt: does this mean we're not storing the individual fields, just the concatenated address?

PR #497, opened 2019-08-30
HTBHF-2189 Added error response from claimant service to Error when thrown

YetAnotherMatt:
Also renamed `transofrmResponse` to `handleErrorResponse`

### DIFF
--- a/src/web/routes/application/terms-and-conditions/post.js
+++ b/src/web/routes/application/terms-and-conditions/post.js
@@ -14,7 +14,7 @@ const { render } = require('./get')
 
 const handleErrorResponse = (body, response) => {
   if (isErrorStatusCode(response.statusCode)) {
-    throw new Error(`Error posting to claimant service: ${JSON.stringify(body)}`)
+    throw new Error(JSON.stringify(body))
   }
 
   return response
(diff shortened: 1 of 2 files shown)


src/web/routes/application/terms-and-conditions/post.js
```diff
@@ -12,9 +12,9 @@ const { isErrorStatusCode } = require('./predicates')
 const { CLAIMS_ENDPOINT, NO_ELIGIBILITY_STATUS_MESSAGE } = require('./constants')
 const { render } = require('./get')
 
-const transformResponse = (body, response) => {
+const handleErrorResponse = (body, response) => {
   if (isErrorStatusCode(response.statusCode)) {
-    throw new Error('Error posting to claimant service')
+    throw new Error(JSON.stringify(body))
```
mikeSmith08: Wouldn't this be better as `throw new Error('Error posting to claimant service: ' + JSON.stringify(body))` to retain the context of when the error occurs?

mikeSmith08: Also, could we output the HTTP response code too, that could also be useful.

gethyn1: The error is caught and wrapped with "Error posting to claimant service" on line 67. This removes a double occurrence of "Error posting to claimant service"

PR #499, opened 2019-09-02
Map OS places API response to claimant service address format

gethyn1:
- Moved OS places address keys to common constants file
- Map address returned from OS places API to format required by claimant service (N.B. assumes address has been normalised by /postcode adapter):

1. Normalise address by setting any missing fields to a default value of an empty string
2. Construct the 4 possible address parts for address lines 1 and 2:
   a) Organisation name
   b) Building name
   c) Number and street
   d) Locality
3. Split a single part address into two parts if a comma exists in the first part
4. Convert a multipart address into two parts by joining all parts except the first with a comma


src/web/routes/application/select-address/adapters.test.js
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  const expected = 'HAWARDEN INDUSTRIAL PARK, PENARLAG'
+  const result = constructLocality(fields)
+
+  t.equal(result, expected, 'correctly constructs locality')
+  t.end()
+})
+
+test('constructNumberAndStreet() correctly constructs locality', (t) => {
```
dwybourn: Copy and paste error.

gethyn1: Updated

src/web/routes/application/select-address/adapters.addresses.test.js
```diff
@@ -0,0 +1,193 @@
+const test = require('tape')
+const { transformAddress } = require('./adapters')
+
+test('transformAddress() builds address A', (t) => {
```
mikeSmith08: I know we don't normally do this, but would it be worthwhile running these examples past Karen before we merge them in, might be worthwhile to double check with her first. I'm happy with them though.

gethyn1: Yes I'm happy to do that, @YetAnotherMatt what do you think (as you were discussing the format with Karen)?

src/web/routes/application/select-address/adapters.js
```diff
@@ -0,0 +1,75 @@
...
+ *    d) Locality
+ * 3. Split a single part address into two parts if a comma exists in the first part
+ * 4. Convert a multipart address into two parts by joining all parts except the first with a comma
+ */
+
+const isNotEmpty = compose(not, isEmpty)
+
+const joinAddressFields = addressKeys => compose(join(DELIMITER), filter(isNotEmpty), props(addressKeys))
+
+const constructThoroughfare = joinAddressFields(THOROUGHFARE_PROPS)
```
mikeSmith08: I found this (and the two below) constant quite confusing, if its not used anywhere else, could we inline them here, I think seeing `joinAddressFields['DEPENDENT_THOROUGHFARE_NAME', 'THOROUGHFARE_NAME']` here would be much clearer.

src/web/routes/application/select-address/adapters.js
```diff
@@ -0,0 +1,75 @@
+const { props, propOr, pickAll, compose, map, join, filter, not, isEmpty, split, head, tail, defaultTo } = require('ramda')
```
mikeSmith08: Contender for most Ramda functions in a single file? :-)

src/web/routes/application/select-address/adapters.addresses.test.js
YetAnotherMatt: Good idea to share them with Karen, though I'm not sure we need to hold up merging the PR - we can always address any changes before the story is completed.

src/web/routes/application/select-address/adapters.js
gethyn1: Sure - I thought it was cleaner to extract these to constants (as the actual prop names seemed irrelevant) but happy to inline

src/web/routes/application/select-address/adapters.js
```diff
@@ -0,0 +1,75 @@
...
+
+const joinAddressFields = addressKeys => compose(join(DELIMITER), filter(isNotEmpty), props(addressKeys))
+
+const constructThoroughfare = joinAddressFields(THOROUGHFARE_PROPS)
+
+const constructBuildingName = joinAddressFields(BUILDING_NAME_PROPS)
+
+const constructLocality = joinAddressFields(LOCALITY_PROPS)
+
+const constructNumberAndStreet = addressFields => `${propOr('', 'BUILDING_NUMBER', addressFields)} ${constructThoroughfare(addressFields)}`.trim()
```
mikeSmith08: This is hard to see in GitHub, is it just doing a string concatenation of the two values (building number and thoroughfare) with a space in-between and then calling trim on the result?

src/web/routes/application/select-address/adapters.addresses.test.js
YetAnotherMatt: It seems I replied in the wrong place for it to show up here...
Good idea to share them with Karen, though I'm not sure we need to hold up merging the PR - we can always address any changes before the story is completed.

src/web/routes/application/select-address/adapters.js
gethyn1: Yes that is correct - I am actually not totally happy with this as it is difficult to read .. but I didn't want to introduce more functions as I think that would add further confusion.

I think ideally the two values would be concatenated with a space and wrapped in a trim function.

src/web/routes/application/select-address/adapters.js
gethyn1: I've moved these constants inline

src/web/routes/application/select-address/adapters.js
```diff
@@ -0,0 +1,76 @@
+const { props, propOr, pickAll, compose, map, join, filter, not, isEmpty, split, head, tail, defaultTo } = require('ramda')
+const { OS_PLACES_ADDRESS_KEYS } = require('../common/constants')
```
YetAnotherMatt: One for another PR perhaps, but could OS_PLACES_ADDRESS_KEYS have constants for each key, and another for all keys? Then we could do:
```
const constructBuildingName = joinAddressFields([OS_PLACES_ADDRESS_KEYS.SUB_BUILDING_NAME, OS_PLACES_ADDRESS_KEYS.BUILDING_NAME])
```
and
```
const normaliseAddressFields = compose(map(defaultTo('')), pickAll(OS_PLACES_ADDRESS_KEYS.ALL))
```

YetAnotherMatt: I realise OS_PLACES_ADDRESS_KEYS.SUB_BUILDING_NAME is a bit wordy

src/web/routes/application/select-address/adapters.js
```diff
@@ -0,0 +1,76 @@
...
+ * 4. Convert a multipart address into two parts by joining all parts except the first with a comma
+ */
+
+const DELIMITER = ', '
+
+const isNotEmpty = compose(not, isEmpty)
+
+const joinAddressFields = addressKeys => compose(join(DELIMITER), filter(isNotEmpty), props(addressKeys))
+
+const constructThoroughfare = joinAddressFields(['DEPENDENT_THOROUGHFARE_NAME', 'THOROUGHFARE_NAME'])
```
YetAnotherMatt: This seems much clearer :) but I worry a bit about using strings rather than constants

gethyn1: Yes I agree, especially as these keys are now shared across postcode and select address.

src/web/routes/application/select-address/adapters.js
gethyn1: I'm not sure I totally understand but sounds like it's a good idea. It feels like this has implications for the postcode page as well so I will merge this PR and then look at creating another PR to address this (I think it also relates to your comment below).

src/web/routes/application/select-address/adapters.js
mikeSmith08: Feel free to use constants for each one, my point was that the constant which represented both of the fields together was confusing.

PR #500, opened 2019-09-02
HTBHF-2167 Added my address is not listed link.

dwybourn:
- Added a link to the manual address page.
- Reordered the manual address and postcode lookup pages as we can not link to a future step.


src/web/views/select-address.njk
```diff
@@ -35,8 +35,11 @@
                 },
                 items: addresses
             }) }}
+
+            <a class="govuk-link govuk-!-font-size-19" href="/manual-address">{{ addressNotListed }}</a>
         {% endif %}
 
+
```
gethyn1: Rogue new line

dwybourn: Fixed.

src/test/a11y/tests/apply.js
```diff
@@ -65,9 +65,19 @@ const apply = [
     })
   },
   {
-    url: URLS['POSTCODE'],
+    url: URLS['MANUAL_ADDRESS'],
```
YetAnotherMatt: Why have we moved manual address before postcode lookup? Not that it makes any functional difference at this point, but in terms of page flow, you should only be able to get to manual address after postcode lookup...

src/test/common/steps/navigation/step-page-actions.js
```diff
@@ -48,6 +48,11 @@ const STEP_PAGE_ACTIONS = [
     page: (pages) => pages.nationalInsuranceNumber,
     actions: async () => enterNinoAndSubmit()
   },
+  {
```
YetAnotherMatt: Same question again here - why is manual address before postcode lookup?

src/test/a11y/tests/apply.js
dwybourn: Because you can't link to a page further in the flow. Our state machine will redirect you.

PR #502, opened 2019-09-02
Feature/address not listed link

dwybourn:
Added a link to go to the manual address page if the user's address is not listed.


src/web/routes/application/select-address/select-address.js
```diff
@@ -18,6 +19,7 @@ const buildAddressOption = result => ({
 
 const behaviourForGet = () => async (req, res, next) => {
   res.locals.addresses = req.session.postcodeLookupResults.map(buildAddressOption)
+  req.session.nextAllowedStep = '/manual-address'
```
dwybourn: Manual address is further in the flow than select-address, therefore this line is needed to prevent the state machine from redirecting the user back to select-address.

gethyn1: We should probably put a comment here explaining why this is necessary.

This also makes me think the paths should be constants somewhere, although not for this PR.

dwybourn: Done.

src/test/acceptance/features/address-lookup.feature
```diff
@@ -20,4 +20,5 @@ Feature: Select address
       When I enter a postcode
       Then I am shown the select address page
       And I am shown a list of addresses
+      And I am shown an address not listed link that navigates to the manual address page
```
mikeSmith08: Does this step need `that navigates to the manual address page`? That should be assertion but I think its superfluous to the step itself.

src/web/routes/application/select-address/select-address.js
mikeSmith08: I don't understand why this is going to the manual address page, shouldn't this skip past it? I guess I don't understand when it should skip the manual page, and when it shouldn't, and where the difference is.

dwybourn: It will skip past it normally, this step allows a user to click on the link that says my address is not listed.
The next subtask is to make sure that the user doesn't see manual-address when they've selected an address.

src/test/acceptance/features/address-lookup.feature
dwybourn: I'm not sure I understand. Are you saying the assertion isn't needed?

YetAnotherMatt: I think Mike' saying that the name of the condition could simply be: `And I am shown an address not listed link`. The assertion should remain the same under the covers

dwybourn: I see, updated the definition to `I am shown an address not listed link`

PR #504, opened 2019-09-02
HTBHF-2167 Updated claim with address from select address page.

dwybourn:
The claim on the session is now updated with the address selected from the select-address page. The address format is the same as that of manual-address.

### DIFF
--- a/src/web/routes/application/select-address/select-address.js
+++ b/src/web/routes/application/select-address/select-address.js
@@ -1,3 +1,5 @@
+const { transformAddress } = require('./adapters')
+
 const pageContent = ({ translate }) => ({
   title: translate('address.title'),
   heading: translate('address.heading'),
@@ -17,10 +19,19 @@ const buildAddressOption = result => ({
   text: result.ADDRESS
 })
 
-const behaviourForGet = () => async (req, res, next) => {
+const behaviourForGet = () => (req, res, next) => {
   res.locals.addresses = req.session.postcodeLookupResults.map(buildAddressOption)
   // Manual address is further in the flow than select-address, therefore this line is needed to prevent the state machine from redirecting the user back to select-address.
   req.session.nextAllowedStep = '/manual-address'
+  next()
+}
+
+const behaviourForPost = () => (req, res, next) => {
+  const address = req.session.postcodeLookupResults.find(result => result.ADDRESS === req.body.selectedAddress)
+  req.session.claim = {
+    ...req.session.claim,
+    ...transformAddress(address)
+  }
   next()
 }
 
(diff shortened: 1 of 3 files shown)


src/web/routes/application/select-address/select-address.js
```diff
@@ -17,22 +19,33 @@ const buildAddressOption = result => ({
...
-const behaviourForGet = () => async (req, res, next) => {
+const behaviourForGet = () => (req, res, next) => {
   res.locals.addresses = req.session.postcodeLookupResults.map(buildAddressOption)
   // Manual address is further in the flow than select-address, therefore this line is needed to prevent the state machine from redirecting the user back to select-address.
   req.session.nextAllowedStep = '/manual-address'
   next()
 }
 
+const behaviourForPost = () => (req, res, next) => {
+  const address = req.session.postcodeLookupResults.find(result => result.ADDRESS === req.body.selectedAddress)
```
mikeSmith08: I think this would be clearer as its own function.

gethyn1: Possibly should explicitly deal with the situation that the address is not found with try / catch (not sure how this would happen)? Would we want to log that as an error?

src/web/routes/application/select-address/select-address.js
```diff
@@ -17,22 +19,33 @@ const buildAddressOption = result => ({
   text: result.ADDRESS
 })
 
-const behaviourForGet = () => async (req, res, next) => {
+const behaviourForGet = () => (req, res, next) => {
   res.locals.addresses = req.session.postcodeLookupResults.map(buildAddressOption)
   // Manual address is further in the flow than select-address, therefore this line is needed to prevent the state machine from redirecting the user back to select-address.
   req.session.nextAllowedStep = '/manual-address'
```
gethyn1: `req.session.nextAllowedStep` happens in a few files .. do you think this should be abstracted into a function? (probably not for this PR)

dwybourn: For such a simply assignment operation I'm not sure it warrants it's own function.

src/web/routes/application/select-address/select-address.js
dwybourn: I've extracted this logic to it's own function which throws an error if no address is found.

src/web/routes/application/select-address/select-address.js
```diff
@@ -17,22 +19,44 @@ const buildAddressOption = result => ({
...
   next()
 }
 
+const findAddress = (selectedAddress, postcodeLookupResults) => {
+  const address = postcodeLookupResults.find(result => result.ADDRESS === selectedAddress)
+
+  if (!address) {
```
gethyn1: `Array.prototype.find` will return `undefined` if an address is not matched. I think this check should be a bit more explicit. There is an `isUndefined` function in the common predicates file.

PR #505, opened 2019-09-03
Feature/skip manual address

dwybourn:
- When selecting an address from the dropdown, the user is no longer shown the manual address page.
- Fixed acceptance running with the address lookup feature enabled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "start": "node src/web",
     "start:dev": "npm run build && nodemon $NODE_DEBUG_OPTION src/web -e js,json,html,njk",
-    "test": "cross-env LOG_LEVEL=silent CLAIMANT_SERVICE_URL=http://localhost:8090 SESSION_DETAILS_PORT=8081 npm-run-all build test:lint test:unit:coverage test:fail_if_test_only_found -p -r start test:session-details test:browser",
+    "test": "cross-env LOG_LEVEL=silent CLAIMANT_SERVICE_URL=http://localhost:8090 SESSION_DETAILS_PORT=8081 OS_PLACES_URI=http://localhost:8090 npm-run-all build test:lint test:unit:coverage test:fail_if_test_only_found -p -r start test:session-details test:browser",
     "test:lint": "standard",
     "test:unit": "tape -r './src/test/unit/setup' 'src/@(web|config)/**/*.test.js'",
     "test:unit:coverage": "nyc npm run test:unit",
--- a/src/test/common/steps/address-lookup-steps.js
+++ b/src/test/common/steps/address-lookup-steps.js
@@ -8,7 +8,7 @@ const { enterPostcodeAndSubmit } = require('./common-steps')
 const { setupPostcodeLookupWithNoResults, setupPostcodeLookupWithResults } = require('../wiremock')
 const { POSTCODE } = require('./constants')
 
-const POSTCODE_WITH_NO_RESULTS = 'BS1 1AA'
+const POSTCODE_WITH_NO_RESULTS = 'BS11AA'
 
 When(/^I enter a postcode with no search results$/, async function () {
   await setupPostcodeLookupWithNoResults(POSTCODE_WITH_NO_RESULTS)
--- a/src/web/routes/application/manual-address/manual-address.js
+++ b/src/web/routes/application/manual-address/manual-address.js
@@ -1,4 +1,4 @@
-const { join, filter, compose } = require('ramda')
+const { join, filter, compose, isNil } = require('ramda')
 const { validate } = require('./validate')
 const { sanitize } = require('./sanitize')
 const { notIsNilOrEmpty } = require('../../../../common/predicates')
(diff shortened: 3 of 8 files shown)


src/test/common/steps/address-lookup-steps.js
```diff
@@ -8,7 +8,7 @@ const { enterPostcodeAndSubmit } = require('./common-steps')
 const { setupPostcodeLookupWithNoResults, setupPostcodeLookupWithResults } = require('../wiremock')
 const { POSTCODE } = require('./constants')
 
-const POSTCODE_WITH_NO_RESULTS = 'BS1 1AA'
+const POSTCODE_WITH_NO_RESULTS = 'BS11AA'
```
dwybourn: Previous normalisation PR caused these tests to fail as the postcode sent (BS11AA) no longer matched that of the wiremock stub (BSA 1AA)

src/web/routes/application/select-address/select-address.js
```diff
@@ -37,11 +39,14 @@ const findAddress = (selectedAddress, postcodeLookupResults) => {
 }
 
 const behaviourForPost = () => (req, res, next) => {
-  const address = findAddress(req.body.selectedAddress, req.session.postcodeLookupResults)
-  req.session.claim = {
-    ...req.session.claim,
-    ...transformAddress(address)
+  if (req.body.selectedAddress) {
```
dwybourn: Bugfix, only transform the address if one has been selected. Can move to own pr if needed.

package.json
```diff
@@ -10,21 +10,21 @@
...
-    "test:compatibility:local:debug": "mkdir -p build/reports/compatibility && cross-env SESSION_DETAILS_PORT=8081 OS_PLACES_URI=http://localhost:8090 TESTS=localCompatibility node --inspect-brk=9229 $NODE_DEBUG_OPTION node_modules/cucumber/bin/cucumber-js src/test/compatibility/features/ --require 'src/test/common/steps/**/*.js' --require 'src/test/common/timeout.js' --require 'src/test/compatibility/config/environment.js' -f json:build/reports/compatibility-report.json",
+    "test:compatibility:local:test": "mkdir -p build/reports/compatibility && cross-env SESSION_DETAILS_PORT=8081 OS_PLACES_URI=http://localhost:8090 OS_PLACES_URI=http://localhost:8090 TESTS=localCompatibility cucumber-js src/test/compatibility/features/ --require 'src/test/common/steps/**/*.js' --require 'src/test/common/timeout.js' --require 'src/test/compatibility/config/environment.js' -f json:build/reports/compatibility-report.json",
```
YetAnotherMatt: Duplicate OS_PLACES_URI in test:compatibility:local:test, test:acceptance:debug and test:compatibility:local:debug

src/web/routes/application/manual-address/manual-address.js
```diff
@@ -32,13 +32,18 @@ const contentSummary = (req) => ({
   ])
 })
 
+const isNavigable = (session) => {
+  return isNil(session.claim.selectedAddress)
```
YetAnotherMatt: I'm not sure this is going to work. What if someone selects an address in the dropdown, then decides that it's not the correct address and clicks 'my address is not listed'?

dwybourn: This is handled in the behaviourForGet

package.json
dwybourn: Removed duplicates.

src/web/routes/application/select-address/select-address.js
```diff
@@ -20,6 +20,8 @@ const buildAddressOption = result => ({
 })
 
 const behaviourForGet = () => (req, res, next) => {
+  // reset this value as this is used to determine whether or not to skip the manual address page
+  delete req.session.claim.selectedAddress
```
mikeSmith08: I'm not sure I understand why this is needed, lets discuss when you get in.

src/web/routes/application/select-address/select-address.test.js
```diff
@@ -159,7 +200,7 @@ test('findAddress throws an error when the address is not found', (t) => {
 
   t.throws(
     findAddress.bind(null, selectedAddress, postcodeLookupResults),
-    'Unable to find selected address in list of postcode lookup results',
+    `Unable to find ${selectedAddress} in list of addresses: ${postcodeLookupResults}`,
```
gethyn1: This test is incorrect as this is not the error that is thrown. I think this test is passing because this argument should be a regex

src/web/routes/application/select-address/select-address.js
gethyn1: Might be worth abstracting this to a named function to clarify the purpose. Something like:
```
const resetAddressState = (req) => {
  if (req.session.claim.selectedAddress) {
    delete req.session.claim.selectedAddress
  }
}
```

gethyn1: Which could also be unit tested

src/web/routes/application/select-address/select-address.test.js
dwybourn: Updated to use a regex.

src/web/routes/application/select-address/select-address.js
dwybourn: Moved into own function and updated behaviourForGet unit test.

src/web/routes/application/select-address/select-address.test.js
```diff
@@ -159,7 +203,7 @@ test('findAddress throws an error when the address is not found', (t) => {
 
   t.throws(
     findAddress.bind(null, selectedAddress, postcodeLookupResults),
-    'Unable to find selected address in list of postcode lookup results',
+    /Unable to find selected address in list of postcode lookup results/,
```
mikeSmith08: Minor - shouldn't this (and the next line) be indented?

PR #506, opened 2019-09-04
HTBHF-2199 Make a call to GA on successful and failed postcode lookup for audit purposes.

mikeSmith08:


src/web/routes/application/postcode/postcode.js
```diff
@@ -8,6 +8,25 @@ const REQUEST_TIMEOUT = 5000
...
 const standardisePostcode = (postcode) => postcode.toUpperCase().replace(/\s/g, '')
 
+const buildGoogleAnalyticsFullUri = (gaBaseUrl, trackingId, result, ipAddress, numberOfResults, sessionId) =>
+  `${gaBaseUrl}?v=1&tid=${trackingId}&t=event&cid=${sessionId}&ec=AddressLookup&ea=${result}&el=${ipAddress}&ev=${numberOfResults}`
+
+const auditSuccessfulPostcodeLookup = async (config, req, numberOfResults) => auditPostcodeLookup(config, req, 'SuccessfulLookup', numberOfResults)
+
+const auditFailedPostcodeLookup = async (config, req) => auditPostcodeLookup(config, req, 'FailedLookup', 0)
+
+const auditPostcodeLookup = async (config, req, result, numberOfResults) => {
```
YetAnotherMatt: Maybe 'result' could be renamed 'outcome' to make it different to 'numberOfResults'

manifest.yml
```diff
@@ -8,6 +8,7 @@ applications:
     CLAIMANT_SERVICE_URL: http://htbhf-claimant-service((space-suffix)).apps.internal:8080
     NODE_ENV: production
     APP_VERSION: ((app-version))
+    GOOGLE_ANALYTICS_URI: https://www.google-analytics.com/collect
```
gethyn1: Is this the only Google Analytics URI we could use? If we decided to track events, would that be the same URI?

src/web/routes/application/postcode/postcode.js
```diff
@@ -8,6 +8,25 @@ const REQUEST_TIMEOUT = 5000
 
 const standardisePostcode = (postcode) => postcode.toUpperCase().replace(/\s/g, '')
 
+const buildGoogleAnalyticsFullUri = (gaBaseUrl, trackingId, result, ipAddress, numberOfResults, sessionId) =>
+  `${gaBaseUrl}?v=1&tid=${trackingId}&t=event&cid=${sessionId}&ec=AddressLookup&ea=${result}&el=${ipAddress}&ev=${numberOfResults}`
+
+const auditSuccessfulPostcodeLookup = async (config, req, numberOfResults) => auditPostcodeLookup(config, req, 'SuccessfulLookup', numberOfResults)
```
gethyn1: Do any of these functions need to be `async` as the code doesn't `await` the result?

mikeSmith08: Looks like you're right, have removed them and all works well.

src/web/routes/application/postcode/postcode.js
mikeSmith08: Done

PR #513, opened 2019-09-04
HTBHF-2167 Fixes a bug where the application would crash on the manua…

dwybourn:
- Using path to get selectedAddress from the session.
- Fixes a bug where refreshing the phone number page would crash if there's no session due to the isNavigable method on manual-address failing.

### DIFF
--- a/src/web/routes/application/manual-address/manual-address.js
+++ b/src/web/routes/application/manual-address/manual-address.js
@@ -1,4 +1,4 @@
-const { join, filter, compose, isNil } = require('ramda')
+const { join, filter, compose, isNil, path } = require('ramda')
 const { validate } = require('./validate')
 const { sanitize } = require('./sanitize')
 const { notIsNilOrEmpty } = require('../../../../common/predicates')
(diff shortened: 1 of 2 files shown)


src/web/routes/application/manual-address/maunal-address.test.js
```diff
@@ -73,3 +73,34 @@ test('Address contentSummary() should return content summary in correct format w
   t.deepEqual(result, expected, 'should return content summary in correct format with address line 2 undefined')
   t.end()
 })
+
+test('isNavigable() returns true when there`s no session', (t) => {
```
gethyn1: Is ``` the correct type of apostrophe for `there`s`? I think it should be `there’s` or `there is`.

dwybourn: I thought that was only the case for user content? Either way, I've updated to 'there is'.

src/web/routes/application/manual-address/manual-address.js
```diff
@@ -33,7 +33,7 @@ const contentSummary = (req) => ({
 })
 
 const isNavigable = (session) => {
```
gethyn1: Suggestion (not PR blocking): as you are essentially composing Ramda functions this could be written as:
```
const isNavigable = compose(isNil, path(['claim', 'selectedAddress']))
```

dwybourn: Done.

PR #514, opened 2019-09-05
HTBHF-2202 Added checking of the exact details sent to the claimant service

YetAnotherMatt:


src/test/common/expected-claim.js
```diff
@@ -0,0 +1,24 @@
+const { isNil } = require('ramda')
+
+const expectedClaim = {}
```
gethyn1: Would we gain anything by making this module a class? It feels a bit odd mutating the `expectedClaim` constant. It might possibly be better to create an instance of `ExpectedClaim` and then call `expectedClaim.setProp(key, value)` (or something similar).

PR #515, opened 2019-09-05
Feature/change address

dwybourn:
- Added acceptance test for changing your address when you've used address


src/test/acceptance/features/address-lookup.feature
```diff
@@ -37,3 +37,10 @@ Feature: Select address
     Given I have entered my details up to the select address page
     When I click the address not listed link
     Then I am shown the manual address page
+
+  # TODO DW HTBHF-2037 include test once address lookup is enabled
+  @ignore
+  Scenario: Changing details for a selected address takes me to the select address page
+    Given I have entered my details up to the check details page and selected an address
```
mikeSmith08: The term `selected an address` isn't clear enough, we should probably mention postcode lookup here otherwise its ambiguous.

src/test/common/steps/address-lookup-steps.js
```diff
@@ -1,15 +1,25 @@
...
 const { POSTCODE } = require('./constants')
 
 const POSTCODE_WITH_NO_RESULTS = 'BS11AA'
 
+const stepActionsWithoutManualAddress = (stepPageActions) => stepPageActions.filter(pageAction => pageAction.page(pages) !== pages.manualAddress)
+
+Given(/^I have entered my details up to the check details page and selected an address$/, async function () {
+  const actionOptions = { ...DEFAULT_ACTION_OPTIONS, selectAddress: true }
+  // manual address page is skipped when the user selects an address
+  const stepActions = stepActionsWithoutManualAddress(STEP_PAGE_ACTIONS)
```
mikeSmith08: I think that these step actions should be somewhere more central so that they can be reused if required, possibly along with `step-page-actions.js` or in the same folder so we can collect all the variants together.

mikeSmith08:
> * Added acceptance test for changing your address when you've used address

This sentence doesn't make sense.

src/test/common/steps/address-lookup-steps.js
dwybourn: We discussed this but decided against it as they're only used in this one place. They're specific to address lookup.

src/test/acceptance/features/address-lookup.feature
dwybourn: There's only one way to get to select address, so can't see how it's ambiguous.

dwybourn:
> > * Added acceptance test for changing your address when you've used address
> 
> This sentence doesn't make sense.

Should be 'you've used address lookup'. Missing the last word.

PR #518, opened 2019-09-05
HTBHF-2203 Converted addresses from OS Places to title case

YetAnotherMatt:
Also removed postcode from the address summary.

### DIFF
--- a/src/web/routes/application/address/postcode/adapters.js
+++ b/src/web/routes/application/address/postcode/adapters.js
@@ -2,14 +2,31 @@ const { compose, map, pick, prop, propOr } = require('ramda')
 const { OS_PLACES_ADDRESS_KEYS } = require('../constants')
 
 const RESULTS_PROP = 'results'
-
 const DELIVERY_POINT_ADDRESS_PROP = 'DPA'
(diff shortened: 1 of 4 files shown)


src/web/routes/application/address/postcode/adapters.js
```diff
@@ -2,14 +2,31 @@ const { compose, map, pick, prop, propOr } = require('ramda')
...
+
+const toTitleCase = (str) => {
+  return str.replace(
+    /\b\w+/g,
+    function (txt) { return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase() }
+  )
+}
+
+const convertCase = (address) => {
+  const titleCase = compose(map(toTitleCase), pick(OS_PLACES_TITLE_CASE_KEYS))(address)
```
gethyn1: As this uses `compose` I would move it outside of this function and then just call it here:

```
// Probably want a better name than this!!
const titleCaseAddress = compose(map(toTitleCase), pick(OS_PLACES_TITLE_CASE_KEYS))

const convertCase = (address) => {
  const titleCase = titleCaseAddress(address)
  ... 
}
```

PR #521, opened 2019-09-06
Feature/address content summary

dwybourn:
- Updated the address content function for select address and manual address so that only one of them is visible on the check details page.
- The change address link on the check details page will now go to select address or manual address, depending on how the user entered an address.

### DIFF
--- a/src/web/routes/application/address/content-summary.js
+++ b/src/web/routes/application/address/content-summary.js
@@ -1,0 +1,21 @@
+const { join, filter, compose } = require('ramda')
+
+const { notIsNilOrEmpty } = require('../../../../common/predicates')
+
+const newLineChar = '\n'
+const toMultiLineString = compose(join(newLineChar), filter(notIsNilOrEmpty))
+
+const addressContentSummary = (req) => ({
+  key: req.t('address.summaryKey'),
+  value: toMultiLineString([
+    req.session.claim.addressLine1,
+    req.session.claim.addressLine2,
+    req.session.claim.townOrCity,
+    req.session.claim.county,
+    req.session.claim.postcode
+  ])
+})
+
+module.exports = {
+  addressContentSummary
+}
--- a/src/web/routes/application/address/manual-address/manual-address.js
+++ b/src/web/routes/application/address/manual-address/manual-address.js
@@ -1,7 +1,7 @@
-const { join, filter, compose, isNil, path } = require('ramda')
+const { addressContentSummary } = require('../content-summary')
+const { compose, isNil, path } = require('ramda')
 const { validate } = require('./validate')
 const { sanitize } = require('./sanitize')
-const { notIsNilOrEmpty } = require('../../../../../common/predicates')
 
 const pageContent = ({ translate }) => ({
   title: translate('address.title'),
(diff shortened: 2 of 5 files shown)


src/web/routes/application/address/manual-address/manual-address.js
```diff
@@ -18,19 +18,11 @@ const pageContent = ({ translate }) => ({
...
-    req.session.claim.townOrCity,
-    req.session.claim.county,
-    req.session.claim.postcode
-  ])
-})
+const contentSummary = (req) => {
+  if (isNavigable(req.session)) {
+    return addressContentSummary(req)
+  }
+}
```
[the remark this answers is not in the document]
dwybourn: Updated for manual address and select address.

src/web/routes/application/address/content-summary.js
```diff
@@ -0,0 +1,21 @@
+const { join, filter, compose } = require('ramda')
+
+const { notIsNilOrEmpty } = require('../../../../common/predicates')
+
+const newLineChar = '\n'
```
mikeSmith08: Can/should this be platform dependant?

src/web/routes/application/address/content-summary.js
```diff
@@ -0,0 +1,21 @@
...
+const toMultiLineString = compose(join(newLineChar), filter(notIsNilOrEmpty))
+
+const addressContentSummary = (req) => ({
```
mikeSmith08: Does this need a unit test?

src/web/routes/application/address/manual-address/maunal-address.test.js
```diff
@@ -62,6 +62,15 @@ test('Address contentSummary() should return content summary in correct format w
   t.end()
 })
 
+test('Address contentSummary() is null when there is a selected address on the session', (t) => {
+  const testReq = assocPath(['session', 'claim', 'selectedAddress'], 'test address', req)
```
mikeSmith08: Could we just have this as JSON rather than using assocPath for clarity? We do this in other unit tests...

src/web/routes/application/address/select-address/select-address.js
```diff
@@ -55,18 +58,22 @@ const behaviourForPost = () => (req, res, next) => {
   next()
 }
 
+const contentSummary = req => path(['session', 'claim', 'selectedAddress'], req) ? addressContentSummary(req) : null
```
mikeSmith08: There seems to be an inconsistency with how you've set the contentSummary in the two pages, could/should we be consistent and use `isNavigable` for both? This line confused me...

src/web/routes/application/address/content-summary.js
dwybourn: Doesn't seem worth the effort as it'll be running on the paas.

src/web/routes/application/address/manual-address/maunal-address.test.js
dwybourn: This is consistent with the rest of the test case. If we make this change it should be it's own PR due to the size of the change.

src/web/routes/application/address/select-address/select-address.js
dwybourn: There is no isNavigable function for this page. It's always shown if it's toggled on.

src/web/routes/application/address/content-summary.js
dwybourn: Added unit test.

PR #525, opened 2019-09-06
HTBHF-2037 set state to in progress when changing postcode

gethyn1:
Reseting application state to IN_PROGRESS when a user selects to change their answer to address. This allows the user to progress to subsequent address pages instead of being redirected back to the “Check answers” page.


src/test/acceptance/features/address-lookup.feature
```diff
@@ -44,3 +44,20 @@ Feature: Select address
     Given I have entered my details up to the check details page and selected an address
     When I choose to change my address
     Then I am shown the select address page
+
+  # TODO DW HTBHF-2037 include test once address lookup is enabled
+  @ignore
+  Scenario: Entering a new postcode when I have selected to change my answer for select address takes me to the select address page
```
mikeSmith08: How about just: `Entering a new postcode when I have selected to change my address takes me to the select address page`.

PR #529, opened 2019-09-09
HTBHF-2227 Added validation to postcode page.

dwybourn:
Added postcode validation and not calling os places if there validation errors.


src/test/acceptance/features/address-lookup.feature
```diff
@@ -61,3 +61,16 @@ Feature: Select address
     When I choose to change my address
     And I click the address not listed link
     Then I am shown the manual address page
+
+  # TODO DW HTBHF-2037 include test once address lookup is enabled
+  @ignore
+  Scenario Outline: Enter an invalid postcode on the postcode lookup page
+    Given I have entered my details up to the postcode page
+    When I enter in <postcode> as my postcode
```
gethyn1: Not a blocker but probably don't need `in` here:
`I enter in <postcode> as my postcode` -> `I enter <postcode> as my postcode`

src/web/routes/application/address/postcode/postcode.js
```diff
@@ -30,6 +34,10 @@ const auditPostcodeLookup = (config, req, outcome, numberOfResults) => {
 }
 
 const behaviourForPost = (config) => async (req, res, next) => {
+  if (!validationResult(req).isEmpty()) {
+    return next()
+  }
+
```
gethyn1: It would be good to split this file up in a subsequent PR - it's slowly growing!!

dwybourn: You'll get test failures without it.

src/web/routes/application/address/postcode/postcode.test.js
```diff
@@ -126,6 +129,19 @@ test('behaviourForPost() handles address lookup error', async (t) => {
   t.end()
 })
 
+test('behaviourForPost() does not call os places when there are validation errors', (t) => {
+  const req = {}
+  const res = {}
+  const next = () => {}
+  request.reset()
+  validationResult.onCall().returns(['error'])
```
gethyn1: Do you need `onCall` here? I thought it could just be:
```
validationResult.returns(['error'])
```
... but I might be wrong.

src/web/routes/application/address/postcode/postcode.js
gethyn1: Yes sure, I didn't mean this part .. it was a general observation on the file. I think the OS places logging could be extracted.

src/web/routes/application/address/postcode/postcode.test.js
dwybourn: You're right, it's not needed.

src/test/acceptance/features/address-lookup.feature
dwybourn: Updated to remove `in`.

PR #533, opened 2019-09-09
Feature/handle os places error

dwybourn:
Acceptance test for handling errors from os places.


src/test/acceptance/features/address-lookup.feature
```diff
@@ -74,3 +74,25 @@ Feature: Select address
       | AA1122BB |
       | A        |
       | 11AA21   |
+
+  # TODO DW HTBHF-2037 include test once address lookup is enabled
+  @ignore
+  Scenario: I am shown an error response when a bad respoonse is returned from os places and I can then enter my address manually
```
YetAnotherMatt: respoonse

src/test/acceptance/features/address-lookup.feature
```diff
@@ -74,3 +74,25 @@ Feature: Select address
...
+    Given I have entered my details up to the postcode page
+    And OS places returns an error response
+    When I enter my postcode
+    Then I am shown the select address page
+    And I am informed that there's a problem with the postcode finder
```
mikeSmith08: Shouldn't this be `And I am informed that there's a problem with the address lookup`?

src/test/acceptance/features/address-lookup.feature
dwybourn: Fixed.

src/test/acceptance/features/address-lookup.feature
dwybourn: Updated to address lookup. I think we initially went for postcode finder because that's what the error text says.

PR #544, opened 2019-09-11
HTBHF-2256 handle 400 response from OS places API

gethyn1:
Handle a 400 response from OS places API:

- set the postcode lookup results to an empty array
- log an invalid postcode lookup to Google Analytics


src/web/routes/application/address/postcode/postcode.js
```diff
@@ -20,6 +20,8 @@ const auditSuccessfulPostcodeLookup = (config, req, numberOfResults) => auditPos
 
 const auditFailedPostcodeLookup = (config, req) => auditPostcodeLookup(config, req, 'FailedLookup', 0)
 
+const auditInvalidPostcodeLookup = (config, req) => auditPostcodeLookup(config, req, 'InvalidLookup', 0)
```
YetAnotherMatt: Shouldn't this be 'InvalidPostcode'?

gethyn1: Updated to `InvalidPostcode`

PR #560, opened 2019-09-19
HTBHF-2276 reset address in claim on postcode lookup

gethyn1:
Reset address in the session to ensure old addresses are cleared out when a new address lookup is performed.

EDIT: Extracted address fields to constants to protect against new address fields not being reset.


src/web/routes/application/address/postcode/os-places.js
```diff
@@ -39,8 +39,6 @@ const auditPostcodeLookup = (config, req, outcome, numberOfResults) => {
   }).catch(reason => logger.error('Failed to call Google Analytics to audit postcode: ', reason))
 }
 
-// TODO DW HTBHF-2177 Move call to os places lookup into this file.
-
```
gethyn1: Call to OS Places API is already in this file, so removing this comment

src/web/routes/application/address/postcode/postcode.js
```diff
@@ -8,11 +8,19 @@ const { stateMachine, states } = require('../../common/state-machine')
 const { validate } = require('./validate')
 const { sanitize } = require('../sanitize')
 
+const ADDRESS_KEYS = ['addressLine1', 'addressLine2', 'townOrCity', 'county', 'postcode']
```
YetAnotherMatt: Can we make these constants, and put them in a common address file?

PR #561, opened 2019-09-19
HTBHF-2297 add timing log to OS Places API request

gethyn1:


src/web/routes/application/address/postcode/os-places.js
```diff
@@ -10,12 +10,18 @@ const standardisePostcode = (postcode) => postcode.toUpperCase().replace(/\s/g,
 const getAddressLookupResults = async (config, postcode) => {
```
YetAnotherMatt: Does getAddressLookupResults need to by async? Don't we want it to be synchronous?

src/web/routes/application/address/postcode/os-places.js
```diff
@@ -10,12 +10,18 @@ const standardisePostcode = (postcode) => postcode.toUpperCase().replace(/\s/g,
...
 
-  return request({
+  const response = await request({
     uri: `${OS_PLACES_URI}${OS_PLACES_API_PATH}?postcode=${standardisedPostcode}&key=${OS_PLACES_API_KEY}`,
     json: true,
     timeout: REQUEST_TIMEOUT
   })
+
+  const requestEndDate = moment()
+  const requestTimeInSeconds = requestEndDate.diff(requestStartDate, 'seconds', true)
```
YetAnotherMatt: Trivial, but could we call these requesStartTime & EndTime, not date?

gethyn1: Updated

PR #562, opened 2019-09-20
HTBHF-2297 minimise redundant address transforms

gethyn1:
Minimise the number of address transforms performed during postcode lookup by:

- Only transforming the single line address field of each result from OS Places API response to title case (instead of all address fields)
- Only converting the full address of the selected address to title case (instead of all addresses)

### DIFF
--- a/src/web/routes/application/address/formats.js
+++ b/src/web/routes/application/address/formats.js
@@ -1,0 +1,15 @@
+const { isString } = require('../../../../common/predicates')
+
+const SINGLE_WORD_REGEX = /\b\w+/g
+
+const toTitleCase = (str) => {
+  return !isString(str) ? str : str.replace(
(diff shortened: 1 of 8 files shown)


src/web/routes/application/address/formats.js
```diff
@@ -0,0 +1,15 @@
+const { isString } = require('../../../../common/predicates')
+
+const SINGLE_WORD_REGEX = /\b\w+/g
+
+const toTitleCase = (str) => {
+  return !isString(str) ? str : str.replace(
+    SINGLE_WORD_REGEX,
+    function (txt) { return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase() }
```
mikeSmith08: Might be worth a couple of tests where the String is 1 char long and 0 chars long.

PR #566, opened 2019-09-26
No-JIRA use library email validation

gethyn1:
The express validation library we implement wraps validator JS which implements an email validation check. It seems better to use an actively maintained validator rather than trying to maintain our own regex (unoptimised regex can have significant performance implications for node).

- Replace custom regex email validation with validator from validation library
- Remove explicit max length validation as email validation implicitly checks for length
- Add unit tests for validation middleware
- Add translation function to express validation test helper function

### DIFF
--- a/src/web/routes/application/common/test/apply-express-validation.js
+++ b/src/web/routes/application/common/test/apply-express-validation.js
@@ -26,6 +26,9 @@ const applyExpressValidation = async (req, middleware) => {
     locals: {}
   }
 
+  // Ensure translation function is applied to request object
+  req.t = str => str
+
   try {
     await callMiddlewareQueue(req, res, middleware)
     return validationResult(req)
--- a/src/web/routes/application/email-address/validate.js
+++ b/src/web/routes/application/email-address/validate.js
@@ -1,21 +1,15 @@
 const { check } = require('express-validator')
 const { translateValidationMessage } = require('../common/translate-validation-message')
 
-// See https://stackoverflow.com/questions/201323/how-to-validate-an-email-address-using-a-regular-expression for regex explanation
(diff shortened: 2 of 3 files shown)


src/web/routes/application/common/test/apply-express-validation.js
```diff
@@ -26,6 +26,9 @@ const applyExpressValidation = async (req, middleware) => {
     locals: {}
   }
 
+  // Ensure translation function is applied to request object
+  req.t = str => str
+
```
gethyn1: This express validation test helper is specific to our app setup. Adding translation function to request as it is required for the tests to run correctly - this stops it being omitted by accident.

src/web/routes/application/email-address/validate.js
```diff
@@ -1,21 +1,15 @@
 const { check } = require('express-validator')
 const { translateValidationMessage } = require('../common/translate-validation-message')
 
-// See https://stackoverflow.com/questions/201323/how-to-validate-an-email-address-using-a-regular-expression for regex explanation
-const EMAIL_ADDRESS_REGEX = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
 const MAX_EMAIL_ADDRESS_LENGTH = 256
```
YetAnotherMatt: Suggest we move this to test code

gethyn1: Updated

PR #568, opened 2019-09-30
HTBHF-2355 update GOV.UK frontend to version 3.2.0

gethyn1:
Updated to version 3 of `govuk-frontend`. Quite a few changes:

- Update `govuk-frontend` and `node-sass` (might need to update `lib-sass` bindings locally after merge)
- Update SASS variables
- Update path references for Nunjucks macros
- Update path references for assets (JS / CSS)
- Update Nunjucks start button macro to use new api


package.json
```diff
@@ -72,15 +72,15 @@
     "express": "^4.16.4",
     "express-session": "^1.15.6",
     "express-validator": "^6.0.0",
-    "govuk-frontend": "^2.12.0",
+    "govuk-frontend": "^3.2.0",
     "helmet": "^3.15.1",
```
YetAnotherMatt: Is it worth upgrading helmet at the same time? This is the change in that Snyk pr: 
"helmet": "^3.21.1",

gethyn1: I would like to keep them separate so we can easily roll back the upgrade if things go wrong. The `node-sass` update is related to the upgrade.

PR #579, opened 2019-10-18
Remove unused selenium-js common code, plus acceptance test entries in package.json.

mikeSmith08:
- Remove acceptance test step, page and common files.
 - Remove smoke test feature as we no longer use it (to be double checked but couldn't see any usages).
 - Move constants that were used by a11y tests into the apply.js file and deleted the rest.
 - Tidied up the package.json file to remove acceptance test targets
 - Ran npm uninstall to remove all the unused modules.

There is still more tidying up to do from this which will come from subsequent PRs, mostly around moving stuff into a11y directories and rationalising the remaining environment variables and config files.


src/test/a11y/constants.js
```diff
@@ -0,0 +1,48 @@
+function randomCharFromChars (chars) {
```
mikeSmith08: This file was the remnants of what was needed from constants.js file in the common directory.

PR #590, opened 2019-10-28
Call state machine to set next allowed path

gethyn1:
EPIC: UI support multiple journeys

Route all middleware operations that set the next allowed path in session through the state machine (instead of mutating the session directly). Future PRs will update the state machine to query the user journey when setting this state.

- Create state machine `SET_NEXT_ALLOWED_PATH` action 
- Update middleware session mutations to call state machine

### DIFF
--- a/src/web/routes/application/middleware/handle-path-request/handle-path-request.js
+++ b/src/web/routes/application/middleware/handle-path-request/handle-path-request.js
@@ -3,11 +3,12 @@ const { stateMachine, actions, states } = require('../../common/state-machine')
 const { isPathInApplicationFlow, stepNotNavigable } = require('./predicates')
 const { getPathsInSequence } = require('./selectors')
 
-const { IS_PATH_ALLOWED, GET_NEXT_ALLOWED_PATH } = actions
+const { COMPLETED } = states
+const { IS_PATH_ALLOWED, GET_NEXT_ALLOWED_PATH, SET_NEXT_ALLOWED_PATH } = actions
 
 const middleware = (config, pathsInSequence, step) => (req, res, next) => {
   // Destroy the session on navigating away from CONFIRM_URL
-  if (stateMachine.getState(req) === states.COMPLETED && req.path !== CONFIRM_URL) {
+  if (stateMachine.getState(req) === COMPLETED && req.path !== CONFIRM_URL) {
     req.session.destroy()
     res.clearCookie('lang')
 
@@ -18,13 +19,16 @@ const middleware = (config, pathsInSequence, step) => (req, res, next) => {
     return next()
   }
 
-  // Initialise nextAllowedStep if none exists in session
-  if (!req.session.nextAllowedStep) {
-    req.session.nextAllowedStep = pathsInSequence[0]
+  // Initialise nextAllowedPath if none exists in session
+  let nextAllowedPath = stateMachine.dispatch(GET_NEXT_ALLOWED_PATH, req)
+
+  if (!nextAllowedPath) {
+    const firstPathInSequence = pathsInSequence[0]
+    stateMachine.dispatch(SET_NEXT_ALLOWED_PATH, req, firstPathInSequence)
+    nextAllowedPath = firstPathInSequence
   }
 
   const isPathAllowed = stateMachine.dispatch(IS_PATH_ALLOWED, req, pathsInSequence)
-  const nextAllowedPath = stateMachine.dispatch(GET_NEXT_ALLOWED_PATH, req)
 
   // Redirect to nextAllowedPath on invalid path request
   if (!isPathAllowed) {
--- a/src/web/routes/application/middleware/handle-post.js
+++ b/src/web/routes/application/middleware/handle-post.js
@@ -3,7 +3,7 @@ const { validationResult } = require('express-validator')
 const { wrapError } = require('../common/formatters')
 const { stateMachine, actions } = require('../common/state-machine')
 
-const { GET_NEXT_PATH, INVALIDATE_REVIEW, SET_NEXT_ALLOWED_PATH } = actions
+const { INVALIDATE_REVIEW, INCREMENT_NEXT_ALLOWED_PATH } = actions
 
 const stepInvalidatesReview = (step, claim) => typeof step.shouldInvalidateReview === 'function' && step.shouldInvalidateReview(claim)
 
@@ -26,8 +26,7 @@ const handlePost = (steps, step) => (req, res, next) => {
       stateMachine.dispatch(INVALIDATE_REVIEW, req)
     }
 
-    const nextAllowedPath = stateMachine.dispatch(GET_NEXT_PATH, req, steps)
-    stateMachine.dispatch(SET_NEXT_ALLOWED_PATH, req, nextAllowedPath)
+    stateMachine.dispatch(INCREMENT_NEXT_ALLOWED_PATH, req, steps)
 
     return next()
   } catch (error) {
--- a/src/web/routes/application/middleware/handle-post.js
+++ b/src/web/routes/application/middleware/handle-post.js
@@ -3,7 +3,7 @@ const { validationResult } = require('express-validator')
 const { wrapError } = require('../common/formatters')
 const { stateMachine, actions } = require('../common/state-machine')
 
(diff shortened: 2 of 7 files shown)


src/web/routes/application/middleware/handle-path-request/handle-path-request.js
```diff
@@ -18,13 +19,16 @@ const middleware = (config, pathsInSequence, step) => (req, res, next) => {
     return next()
   }
 
-  // Initialise nextAllowedStep if none exists in session
-  if (!req.session.nextAllowedStep) {
-    req.session.nextAllowedStep = pathsInSequence[0]
+  // Initialise nextAllowedPath if none exists in session
```
YetAnotherMatt: This comment doesn't seem accurate any more?

src/web/routes/application/middleware/handle-path-request/handle-path-request.js
```diff
@@ -3,11 +3,12 @@ const { stateMachine, actions, states } = require('../../common/state-machine')
...
 const { getPathsInSequence } = require('./selectors')
 
-const { IS_PATH_ALLOWED, GET_NEXT_ALLOWED_PATH } = actions
+const { COMPLETED } = states
+const { IS_PATH_ALLOWED, GET_NEXT_ALLOWED_PATH, SET_NEXT_ALLOWED_PATH } = actions
 
 const middleware = (config, pathsInSequence, step) => (req, res, next) => {
   // Destroy the session on navigating away from CONFIRM_URL
-  if (stateMachine.getState(req) === states.COMPLETED && req.path !== CONFIRM_URL) {
+  if (stateMachine.getState(req) === COMPLETED && req.path !== CONFIRM_URL) {
```
mikeSmith08: I think this could be refactored out into its own function for clarity.

src/web/routes/application/middleware/handle-path-request/handle-path-request.js
```diff
@@ -18,13 +19,16 @@ const middleware = (config, pathsInSequence, step) => (req, res, next) => {
     return next()
   }
 
-  // Initialise nextAllowedStep if none exists in session
-  if (!req.session.nextAllowedStep) {
-    req.session.nextAllowedStep = pathsInSequence[0]
+  // Initialise nextAllowedPath if none exists in session
+  const firstPathInSequence = pathsInSequence[0]
```
YetAnotherMatt: This `const` might make more sense inside the if statement below

src/web/routes/application/middleware/handle-path-request/handle-path-request.js
```diff
@@ -18,13 +19,16 @@ const middleware = (config, pathsInSequence, step) => (req, res, next) => {
...
+  let nextAllowedPath = stateMachine.dispatch(GET_NEXT_ALLOWED_PATH, req)
+
+  if (!nextAllowedPath) {
+    stateMachine.dispatch(SET_NEXT_ALLOWED_PATH, req, firstPathInSequence)
+    nextAllowedPath = firstPathInSequence
```
mikeSmith08: Feels like there should be a cleaner way to do this without mutating the nextAllowedPath variable but can't see anything obvious without the code in front of me, have you tried a way without the mutation?

Also, `firstPathInSequence` is only used in the if block so lets move it into there.

src/web/routes/application/middleware/handle-post.js
```diff
@@ -26,7 +26,8 @@ const handlePost = (steps, step) => (req, res, next) => {
       stateMachine.dispatch(INVALIDATE_REVIEW, req)
     }
 
-    req.session.nextAllowedStep = stateMachine.dispatch(GET_NEXT_PATH, req, steps)
+    const nextAllowedPath = stateMachine.dispatch(GET_NEXT_PATH, req, steps)
+    stateMachine.dispatch(SET_NEXT_ALLOWED_PATH, req, nextAllowedPath)
```
mikeSmith08: These two lines are duplicated here and in `terms-and-conditions/post.js` should they be moved as a function into the state machine?

src/web/routes/application/middleware/handle-path-request/handle-path-request.js
gethyn1: Yes, I agree - its possible but involves calling the state machine twice which seemed unnecessary.

src/web/routes/application/middleware/handle-post.js
gethyn1: yes I thought about that .. in the end I decided it was too much of an abstraction; I don't think the state machine should care about an implementation detail such as "set `nextAllowedStep` from the result of calling `getNextPath` for the active state".

This would involve complicating the state machine for the sake of a small amount of repeated code in middleware. I also think that the repeated code in middleware is a lot clearer than abstracting this to the state machine.

Ultimately I think this abstraction would make the code harder to understand - feel free to disagree!!

src/web/routes/application/middleware/handle-path-request/handle-path-request.js
gethyn1: Agreed, but updating functions in this file is not really the focus of this PR. I might do this in a subsequent PR.

src/web/routes/application/middleware/handle-path-request/handle-path-request.js
gethyn1: So the alternative would look something like:
```
const shouldInitialiseNextAllowedPath = (req) => isUndefined(stateMachine.dispatch(GET_NEXT_ALLOWED_PATH, req))

...

// Initialise nextAllowedPath if none exists in session
if (shouldInitialiseNextAllowedPath(req)) {
  stateMachine.dispatch(SET_NEXT_ALLOWED_PATH, req, pathsInSequence[0])
}

const nextAllowedPath = stateMachine.dispatch(GET_NEXT_ALLOWED_PATH, req)
const isPathAllowed = stateMachine.dispatch(IS_PATH_ALLOWED, req, pathsInSequence)
```
Another option would be to create a function that initialises the session outside of this middleware, which would negate the need for the `undefined` check.

Any thoughts on those options?

src/web/routes/application/middleware/handle-post.js
gethyn1:
```
// State machine
...
setNextAllowedPathFromNextPath: (req, steps) => {
  const nextAllowedPath = stateMachine.dispatch(GET_NEXT_PATH, req, steps, req.path)
  stateMachine.dispatch(SET_NEXT_ALLOWED_PATH, req, nextAllowedPath)
}
...

// Calling code
stateMachine.dispatch(SET_NEXT_ALLOWED_PATH_FROM_NEXT_PATH, req, steps)
```
.. maybe this is OK, but it feels messy and not sure we gain much except extra unit tests.

src/web/routes/application/middleware/handle-path-request/handle-path-request.js
mikeSmith08: Yes, I see what you mean, and now I see why the initial solution was written as it is. I wouldn't want to call the state machine twice, so I'm happy to leave it as it was initially just with the `firstPathInSequence` variable in the if block.

src/web/routes/application/middleware/handle-post.js
mikeSmith08: As long as you've considered it, then I'm happy to keep it as it is if it looks like an abstraction too far.

gethyn1: Having thought about this again, there's probably a distinction to be made between setting the next allowed path directly and just incrementing (can't think of a better word) the next allowed path to the next path in the sequence.

gethyn1: @mikeSmith08 I've implemented a refactor - can you look at the last commit and see what you think?

src/web/routes/application/middleware/handle-path-request/handle-path-request.js
gethyn1: Updated

src/web/routes/application/middleware/handle-post.js
mikeSmith08: Yep, looks good to me, that's quite tidy.

PR #600, opened 2019-11-04
HTBHF-2436 Updated first and last name labels.

dwybourn:
First and last name fields updated inline with latest user research.


mikeSmith08:
Dave, you shouldn't be doing that as its in Awaiting Development and we haven't done a kick-off for it.

PR #601, opened 2019-11-04
HTBHF-2545 add path sequence to journey object

gethyn1:
Add `pathsInSequence` property to journey objects at app startup. Avoids recalculation of the path sequence in distinct parts of the application (will be used in state machine in future PRs):

- Add paths in sequence to journey on app startup
- Move `getPathsInSequence` to register journey file (originally in selectors.js - this is not new code)
- Remove redundant `handlePathRequest` middleware closure
- Remove redundant wrapping of native JS function (`isPathInApplicationFlow`)
- Add missing unit tests for `stepNotNavigable`. This function confused me when I tried to debug so thought it was worth unit testing


src/web/routes/application/flow-control/middleware/handle-path-request/handle-path-request.js
```diff
@@ -1,18 +1,19 @@
...
+const handleRequestForPath = (config, journey, step) => (req, res, next) => {
+  const { pathsInSequence } = journey
+
   // Destroy the session on navigating away from CONFIRM_URL
   if (stateMachine.getState(req) === COMPLETED && req.path !== CONFIRM_URL) {
     req.session.destroy()
     res.clearCookie('lang')
 
-    if (isPathInApplicationFlow(req.path, pathsInSequence)) {
+    if (pathsInSequence.includes(req.path)) {
```
gethyn1: This seemed like an unnecessary abstraction as the unit tests were testing core javascript. I think the inline code reads better than the abstraction.

PR #605, opened 2019-11-07
HTBHF-2548 create state machine journey based operators

gethyn1:
Create utility functions for getting / setting properties on state for each journey. Will be implemented in the state machine in a future PR.

### DIFF
--- a/src/web/routes/application/flow-control/state-machine/operators.js
+++ b/src/web/routes/application/flow-control/state-machine/operators.js
@@ -1,7 +1,0 @@
-const setNextAllowedPath = (req, journey, path) => {
-  req.session.nextAllowedStep = path
-}
-
-module.exports = {
-  setNextAllowedPath
-}
--- a/src/web/routes/application/flow-control/state-machine/session-accessors.test.js
+++ b/src/web/routes/application/flow-control/state-machine/session-accessors.test.js
@@ -1,0 +1,99 @@
+const test = require('tape')
+const { JOURNEYS_KEY, NEXT_ALLOWED_PATH_KEY, STATE_KEY } = require('./keys')
+const { setNextAllowedPathInSession, setStateInSession, getNextAllowedPathFromSession, getStateFromSession } = require('./session-accessors')
+
+const JOURNEY = { name: 'report-a-change' }
+
+test('setNextAllowedPathInSession() sets the next allowed path for the correct journey in session', (t) => {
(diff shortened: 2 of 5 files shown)


src/web/routes/application/flow-control/state-machine/operators.js
```diff
@@ -1,7 +1,27 @@
+const { JOURNEYS_KEY, NEXT_ALLOWED_PATH_KEY, STATE_KEY } = require('./keys')
```
mikeSmith08: How about calling this file `session-accessors.js`?

src/web/routes/application/flow-control/state-machine/operators.test.js
```diff
@@ -0,0 +1,99 @@
+const test = require('tape')
+const { JOURNEYS_KEY, NEXT_ALLOWED_PATH_KEY, STATE_KEY } = require('./keys')
+const { setNextAllowedPathInSession, setStateInSession, getNextAllowedPathFromSession, getStateFromSession } = require('./operators')
+
+const journey = { name: 'report-a-change' }
```
mikeSmith08: Can we make this upper case so its more obvious in the tests that it a constant in the file? Too me a bit to figure out where `journey` was coming from in the tests.

src/web/routes/application/flow-control/state-machine/operators.js
gethyn1: Good naming, I have updated.

src/web/routes/application/flow-control/state-machine/operators.test.js
gethyn1: Yep.

src/web/routes/application/flow-control/state-machine/session-accessors.test.js
```diff
@@ -0,0 +1,99 @@
+const test = require('tape')
+const { JOURNEYS_KEY, NEXT_ALLOWED_PATH_KEY, STATE_KEY } = require('./keys')
+const { setNextAllowedPathInSession, setStateInSession, getNextAllowedPathFromSession, getStateFromSession } = require('./session-accessors')
+
+const JOURNEY = { name: 'report-a-change' }
```
dwybourn: Could we rename this to REPORT_A_CHANGE_JOURNEY? I think it would make the tests a little bit quicker to read.

PR #606, opened 2019-11-11
Feature/session test utils

gethyn1:
Updating the session shape (for implementing multiple user journeys) requires a number of unit tests maintaining knowledge of the shape of the session. Exposing test utilities from the flow control module aims to abstract the knowledge of that session shape away from individual unit tests.

Implementing these test utils will look something like the following:
```
const getStateForApplyJourney = getStateForJourney('apply')

test('Dispatching INVALIDATE_REVIEW should set state to IN_PROGRESS', async (t) => {
  const req = {
    session: {
      ...buildSessionForJourney({ journeyName: 'apply', state: IN_REVIEW })
    },
    path: '/first'
  }

  stateMachine.dispatch(INVALIDATE_REVIEW, req, APPLY_JOURNEY)
  t.equal(getStateForApplyJourney(req), IN_PROGRESS, 'sets state to IN_PROGRESS')
  t.end()
})
```

### DIFF
--- a/src/web/routes/application/flow-control/state-machine/session-accessors.test.js
+++ b/src/web/routes/application/flow-control/state-machine/session-accessors.test.js
@@ -45,6 +45,23 @@ test('setJourneySessionProp() throws an error if journey name is undefined', (t)
   t.end()
 })
 
+test('getJourneySessionProp() throws an error if journey is undefined', (t) => {
+  const req = {
+    session: {
+      [JOURNEYS_KEY]: {
+        'report-a-change': {
+          nextPath: '/first'
+        }
+      }
+    }
+  }
+
+  const result = () => getJourneySessionProp('nextPath')(req, undefined)
+
+  t.throws(result, /No journey defined when trying to get "nextPath"/, 'throws an error if journey name is undefined')
+  t.end()
+})
+
 test('getJourneySessionProp() throws an error if journey name is undefined', (t) => {
   const req = {
     session: {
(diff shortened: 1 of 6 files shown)


src/web/routes/application/flow-control/state-machine/test-utils/test-utils.js
```diff
@@ -0,0 +1,25 @@
+const { path } = require('ramda')
+const { JOURNEYS_KEY, STATE_KEY, NEXT_ALLOWED_PATH_KEY } = require('../keys')
+
+const buildSessionStatusForJourney = ({ name, state, nextAllowedPath }) => ({
+  [JOURNEYS_KEY]: {
```
gethyn1: Update `name` to `journeyName`

gethyn1: Rename to `buildSessionStatusForJourney`

src/web/routes/application/flow-control/state-machine/session-accessors.test.js
```diff
@@ -45,6 +45,23 @@ test('setJourneySessionProp() throws an error if journey name is undefined', (t)
...
+    session: {
+      [JOURNEYS_KEY]: {
+        'report-a-change': {
+          nextPath: '/first'
+        }
+      }
+    }
+  }
+
+  const result = () => getJourneySessionProp('nextPath')(req)
```
gethyn1: Explicitly add `undefined` argument

PR #607, opened 2019-11-12
HTBHF-2548 get and set journey session state

gethyn1:
This is a breaking change as the shape of the session is being updated. Any active users with active sessions when this update reaches production will encounter errors the next time they make a request from the server:

- Update the state machine to read and write to the correct part of the session depending on the journey associated with the requested route
- Update all code that calls the state machine to pass the journey as an argument
- Update unit tests to reflect the change in session shape

### DIFF
--- a/src/web/routes/application/flow-control/middleware/handle-path-request/handle-path-request.js
+++ b/src/web/routes/application/flow-control/middleware/handle-path-request/handle-path-request.js
@@ -3,27 +3,21 @@ const { stateMachine, actions, states } = require('../../state-machine')
 const { stepNotNavigable } = require('./predicates')
 
 const { COMPLETED } = states
-const { IS_PATH_ALLOWED, GET_NEXT_ALLOWED_PATH, SET_NEXT_ALLOWED_PATH } = actions
+const { IS_PATH_ALLOWED, GET_NEXT_ALLOWED_PATH } = actions
 
+// TODO HTBHF-2506 remove config from signature as no longer required
 const handleRequestForPath = (config, journey, step) => (req, res, next) => {
   const { pathsInSequence } = journey
   const firstPathInSequence = pathsInSequence[0]
 
   // Destroy the session on navigating away from CONFIRM_URL
-  if (stateMachine.getState(req) === COMPLETED && req.path !== CONFIRM_URL) {
+  if (stateMachine.getState(req, journey) === COMPLETED && req.path !== CONFIRM_URL) {
     req.session.destroy()
     res.clearCookie('lang')
     return res.redirect(firstPathInSequence)
   }
 
-  // Initialise nextAllowedPath if none exists in session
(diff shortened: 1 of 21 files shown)


src/web/routes/application/flow-control/middleware/handle-path-request/handle-path-request.js
```diff
@@ -3,27 +3,21 @@ const { stateMachine, actions, states } = require('../../state-machine')
...
-  // Initialise nextAllowedPath if none exists in session
-  let nextAllowedPath = stateMachine.dispatch(GET_NEXT_ALLOWED_PATH, req, journey)
-
-  if (!nextAllowedPath) {
-    stateMachine.dispatch(SET_NEXT_ALLOWED_PATH, req, journey, firstPathInSequence)
-    nextAllowedPath = firstPathInSequence
-  }
-
+  const nextAllowedPath = stateMachine.dispatch(GET_NEXT_ALLOWED_PATH, req, journey)
   const isPathAllowed = stateMachine.dispatch(IS_PATH_ALLOWED, req, journey)
```
gethyn1: This has been removed as the session is now initialised with the first path in the journey sequence, so `GET_NEXT_ALLOWED_PATH` will always return the correct path.

src/web/routes/application/flow-control/middleware/handle-path-request/handle-path-request.test.js
```diff
@@ -15,26 +21,11 @@ const config = {
...
-  const next = sinon.spy()
-
-  handleRequestForPath(config, journey)(req, res, next)
-
-  t.equal(req.session.nextAllowedStep, '/first', 'it should set next allowed path to first in sequence')
-  t.equal(next.called, true, 'it should call next()')
-  t.end()
-})
-
 test('handleRequestForPath() should redirect to next allowed step if requested path is not allowed', (t) => {
```
gethyn1: Removing redundant test (see comment on `handle-path-request.js`)

src/web/routes/application/flow-control/state-machine/test-utils/test-utils.js
```diff
@@ -14,9 +14,9 @@ const JOURNEYS_PATH = ['session', JOURNEYS_KEY]
 
 const getJourneyPath = name => [...JOURNEYS_PATH, name]
 
-const getStateForJourney = name => path([...getJourneyPath(name), STATE_KEY])
+const getStateForJourney = (name, req) => path([...getJourneyPath(name), STATE_KEY], req)
 
-const getNextAllowedPathForJourney = (name) => path([...getJourneyPath(name), NEXT_ALLOWED_PATH_KEY])
+const getNextAllowedPathForJourney = (name, req) => path([...getJourneyPath(name), NEXT_ALLOWED_PATH_KEY], req)
```
gethyn1: Updated the signature of these two functions to make it simpler for calling code

PR #610, opened 2019-11-14
HTBHF-2595 prefix path utility function

gethyn1:
Multiple user journeys requires the ability to prefix constant paths with a journey specific prefix e.g:
```
const CHECK_ANSWERS_URL = '/check-answers'
const JOURNEY_PATH_PREFIX = '/report-a-change'

const applyCheckAnswersUrl = prefixPath(JOURNEY_PATH_PREFIX, CHECK_ANSWERS_URL)
// -> '/report-a-change/check-answers'
```
- Create prefix path utility function
- Move `paths.js` into `paths` module


src/web/routes/application/paths/prefix-path.test.js
```diff
@@ -0,0 +1,24 @@
...
+  t.end()
+})
+
+test('prefixPath() throws an error if prefix does not start with "/"', (t) => {
+  const result = () => prefixPath('my-journey', '/first')
+  t.throws(result, /Invalid prefix "my-journey". Prefix must be a string starting with "\/"/, 'throws an error if prefix does not start with "/"')
+  t.end()
+})
+
+test('prefixPath() throws an error if prefix does not start with "/"', (t) => {
```
YetAnotherMatt: `prefixPath() throws an error if prefix is not a string`?

src/web/routes/application/paths/prefix-path.js
```diff
@@ -0,0 +1,23 @@
+const { isUndefined, isString } = require('../../../../common/predicates')
+
+const isInvalidSegment = segment => !isString(segment) || !segment.startsWith('/')
+
+const prefixPath = (prefix, path) => {
+  if (isInvalidSegment(path)) {
+    throw new Error(`Invalid path "${path}". Path must be a string starting with "/"`)
+  }
+
+  if (isUndefined(prefix)) {
```
dwybourn: Is this check necessary? Won't it be covered by isInvalidSegment(prefix) below?

YetAnotherMatt: No, because this returns the path unchanged. isInvalidSegment throws an error.

PR #611, opened 2019-11-14
HTBHF-2595 prefix constant journey paths

gethyn1:
Update any uses of constant journey URLs (`CHECK_ANSWERS_URL` / `TERMS_AND_CONDITIONS_URL` / `CONFIRM_URL`) to implement `pathPrefix()`. This ensures these URLs are prefixed with `journey.pathPrefix` where required.

As there is no `pathPrefix` property defined on the apply journey this does not impact the routing of the current apply journey:

- Update signature of state machine selector functions to use `journey` instead of `steps` where `journey.pathPrefix` is required
- Implement `pathPrefix` when referencing constant URLs in route handlers / state machine
- Implement `pathPrefix` when registering routes using constant URLs
- Update template references to constant URLs to implement `pathPrefix`

### DIFF
--- a/src/web/routes/application/flow-control/state-machine/selectors/selectors.js
+++ b/src/web/routes/application/flow-control/state-machine/selectors/selectors.js
@@ -1,10 +1,10 @@
-const { CHECK_ANSWERS_URL, TERMS_AND_CONDITIONS_URL } = require('../../../paths')
+const { CHECK_ANSWERS_URL, TERMS_AND_CONDITIONS_URL, prefixPath } = require('../../../paths')
 const { isNextPathNavigable } = require('../predicates')
 const { getNextForStep } = require('./get-next-for-step')
 
 const getStepForPath = (path, steps) => steps.find(step => step.path === path)
 
-const getNextInReviewPath = (req) => req.path === CHECK_ANSWERS_URL ? TERMS_AND_CONDITIONS_URL : CHECK_ANSWERS_URL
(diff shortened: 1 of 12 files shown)


src/web/routes/application/flow-control/state-machine/selectors/selectors.js
```diff
@@ -1,10 +1,10 @@
-const { CHECK_ANSWERS_URL, TERMS_AND_CONDITIONS_URL } = require('../../../paths')
+const { CHECK_ANSWERS_URL, TERMS_AND_CONDITIONS_URL, prefixPath } = require('../../../paths')
```
YetAnotherMatt: maybe rename to something like `addPrefixToPath`? I wasn't immediately sure what `prefixPath` wass.

src/web/routes/application/flow-control/state-machine/selectors/selectors.js
```diff
@@ -1,10 +1,10 @@
...
 const { isNextPathNavigable } = require('../predicates')
 const { getNextForStep } = require('./get-next-for-step')
 
 const getStepForPath = (path, steps) => steps.find(step => step.path === path)
 
-const getNextInReviewPath = (req) => req.path === CHECK_ANSWERS_URL ? TERMS_AND_CONDITIONS_URL : CHECK_ANSWERS_URL
+const getNextInReviewPath = (req, prefix) => req.path === prefixPath(prefix, CHECK_ANSWERS_URL) ? prefixPath(prefix, TERMS_AND_CONDITIONS_URL) : prefixPath(prefix, CHECK_ANSWERS_URL)
```
YetAnotherMatt: Is it worth pulling `prefixPath(prefix, CHECK_ANSWERS_URL)` into a `const`? It's used twice in this ternary.

PR #616, opened 2019-11-18
Feature/os places language

dwybourn:
Postcode lookup now adds a language query param to the os places uri.
The user's language (from their user agent) will be used if it is either 'en' or 'cy', otherwise the default of 'en' will be used.

### DIFF
--- a/src/web/routes/application/steps/address/postcode/os-places.js
+++ b/src/web/routes/application/steps/address/postcode/os-places.js
@@ -1,12 +1,9 @@
 const request = require('request-promise')
 const moment = require('moment')
-const { isNil } = require('ramda')
 const { logger } = require('../../../../../logger')
 
 const REQUEST_TIMEOUT = 5000
 const OS_PLACES_API_PATH = '/places/v1/addresses/postcode'
-const OS_PLACES_LANGUAGES = ['en', 'cy']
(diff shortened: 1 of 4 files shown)


dwybourn:
> Can you confirm what happens when we request the site in Welsh and attempt to look up an address in England?

Updated the PR so the language is hardcoded to 'en'

src/web/routes/application/steps/address/postcode/os-places.js
```diff
@@ -7,13 +7,13 @@ const OS_PLACES_API_PATH = '/places/v1/addresses/postcode'
 
 const standardisePostcode = (postcode) => postcode.toUpperCase().replace(/\s/g, '')
 
-const getAddressLookupResults = async (config, postcode) => {
+const getAddressLookupResults = async (config, postcode, language) => {
```
mikeSmith08: You don't need to pass `language` this into this method.

YetAnotherMatt: Do we need the language parameter now?

dwybourn: That's removed now.

dwybourn: Removed.

PR #618, opened 2019-11-19
HTBHF-2482 Added form description to are you pregnant page.

dwybourn:
- Added a form description to the are you pregnant page.
- The fieldSet options within govukRadios does not support having hint text (hint text is only available for the labels), so the radio buttons were extracted to be wrapped within a govukFieldset block and a htbhfFormDescription is used for the hint. (This is the same pattern we use for manual-address).


src/web/views/are-you-pregnant.njk
```diff
@@ -2,52 +2,60 @@
...
+        fieldNamePrefix: "expectedDeliveryDate",
+        claim: claim,
+        errors: errors,
+        legend: {
+            text: expectedDeliveryDateText
+        },
+        hint: {
+            text: expectedDeliveryDateHint
+        }
+    }) }}
```
gethyn1: Can this indentation be reset?

PR #620, opened 2019-11-19
HTBHF-2510 define POST endpoint on journey definition

gethyn1:
Allow journeys to define the endpoint when making POST request on terms and conditions page:

- Tidy unit tests for terms and conditions `post.js`
- Add unit test for claimant service POST request
- Define POST endpoint on journey definition


src/web/routes/application/journey-definitions.js
```diff
@@ -18,6 +18,7 @@ const {
 
 const APPLY = {
   name: 'apply',
+  endpoint: '/v2/claims',
```
dwybourn: endpoint to me sounds like the endpoint to start the journey. claimsEndpoint might work better?

YetAnotherMatt: or `postEndpoint`?

dwybourn: Still sounds like an endpoint on the UI. claimServiceEndpoint or something like that. Just needs to be clear that this is the path of a down stream service at the end of the journey.

gethyn1: I would rather explain this in the README (there is a story for updating this). I think `endpoint` and a search for where it is used is enough .. unless you feel really strongly about this??

YetAnotherMatt: :man_shrugging:

PR #624, opened 2019-11-20
No-Jira update git ignore file

adeel718:


.gitignore
```diff
@@ -1,3 +1,5 @@
+.gradle/
```
gethyn1: Why do we need to ignore a gradle directory in a node project?

YetAnotherMatt: Good point. We don't use gradle in web-ui so this isn't needed. @adeel718 you can just delete the .gradle directory on your local machine.

PR #631, opened 2019-11-27
HTBHF-2495 remove apply link on guidance page

gethyn1:


src/web/views/guidance/en/how-it-works.njk
```diff
@@ -19,7 +19,7 @@
   </ul>
 
   <h2 class="govuk-heading-l">If you’re already on the scheme</h2>
-  <p class="govuk-body">You’ll no longer get milk and food vouchers. You need to <a class="govuk-link" href="/apply">apply for a Healthy Start</a> money card instead.</p>
```
adeel718: More of a question - does the acceptance tests go to the level of checking content and clicking the link?

gethyn1: Not on the guidance pages (I don't think), but if it is critical to the application we will check that.

gethyn1: We check the titles and "previous" and "next" links on each guidance page

gethyn1: https://github.com/DepartmentOfHealth-htbhf/htbhf-acceptance-tests/blob/master/src/test/java/uk/gov/dhsc/htbhf/steps/GuidanceSteps.java

adeel718: Thanks, I have raised this with Jyothi as we don't check content for some of our services, I think we should continue doing this

PR #636, opened 2019-11-29
HTBHF-2675 Added methods for storing and retrieving additional step-related data from the session

YetAnotherMatt:
render-view.js now adds any additionalData for the current step to res.locals

### DIFF
--- a/src/web/routes/application/flow-control/session-accessors/session-accessors.js
+++ b/src/web/routes/application/flow-control/session-accessors/session-accessors.js
@@ -1,6 +1,6 @@
 const { pathOr, compose, toPairs } = require('ramda')
 const { isUndefined } = require('../../../../../common/predicates')
-const { JOURNEYS_KEY, NEXT_ALLOWED_PATH_KEY, STATE_KEY } = require('../keys')
+const { JOURNEYS_KEY, NEXT_ALLOWED_PATH_KEY, STATE_KEY, ADDITIONAL_DATA_KEY } = require('../keys')
 
 const JOURNEYS_PATH = ['session', JOURNEYS_KEY]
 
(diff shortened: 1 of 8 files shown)


src/web/routes/application/flow-control/session-accessors/session-accessors.js
```diff
@@ -41,12 +41,22 @@ const getStateFromSession = getJourneySessionProp(STATE_KEY)
 
 const getJourneysFromSession = compose(toPairs, pathOr({}, JOURNEYS_PATH))
 
+const getAdditionalDataForStep = (req, step) => {
+  return pathOr({}, [ADDITIONAL_DATA_KEY, step.path], req.session)
```
gethyn1: Could add `session` to the path:
```
const getAdditionalDataForStep = (req, step) =>
  pathOr({}, [ADDITIONAL_DATA_KEY, step.path, 'session'], req)
```

src/web/routes/application/flow-control/session-accessors/session-accessors.js
```diff
@@ -41,12 +41,22 @@ const getStateFromSession = getJourneySessionProp(STATE_KEY)
...
+}
+
+const setAdditionalDataForStep = (req, step, stepData) => {
+  req.session[ADDITIONAL_DATA_KEY][step.path] = stepData
```
gethyn1: Should this be set per journey? .. although I guess this might not work if the claim is global.

YetAnotherMatt: To quote @gethyn1: 'no' :)

gethyn1:
Is `stepData` better than `additionalData`? .. discuss.

src/web/routes/application/flow-control/session-accessors/session-accessors.js
YetAnotherMatt: I did start that way, but wasn't quite happy referring to `'session'` when the rest of the code refers to `req.session`

PR #637, opened 2019-12-02
HTBHF-2675 Changed select-address to store address data in session.stepData instead of res.locals

YetAnotherMatt:
Also validating that an address has been selected

### DIFF
--- a/src/web/routes/application/steps/address/select-address/select-address.js
+++ b/src/web/routes/application/steps/address/select-address/select-address.js
@@ -2,7 +2,8 @@ const { path } = require('ramda')
 const { transformAddress } = require('./adapters')
 const { addressContentSummary } = require('../content-summary')
 const { requestBody } = require('../request-body')
-const { stateMachine, actions } = require('../../../flow-control')
+const { stateMachine, actions, setAdditionalDataForStep } = require('../../../flow-control')
+const { validate } = require('./validate')
 
 const { SET_NEXT_ALLOWED_PATH } = actions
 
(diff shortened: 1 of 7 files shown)


src/web/routes/application/steps/address/select-address/select-address.js
```diff
@@ -3,6 +3,8 @@ const { transformAddress } = require('./adapters')
 const { addressContentSummary } = require('../content-summary')
 const { requestBody } = require('../request-body')
 const { stateMachine, actions } = require('../../../flow-control')
+const { validate } = require('./validate')
+const { setAdditionalDataForStep } = require('../../../flow-control/session-accessors')
```
gethyn1: This should be exported from `flow-control` - see line 5

src/web/routes/application/steps/address/select-address/select-address.js
```diff
@@ -33,14 +35,24 @@ const resetAddressState = (req) => {
   }
 }
 
-const behaviourForGet = (config, journey) => (req, res, next) => {
+const getAddressDataFromSession = (req, addressId) => {
+  const addresses = req.session.postcodeLookupResults.map(buildAddressOption(addressId))
```
gethyn1: You could inline this on line 41

YetAnotherMatt: it's also referenced on line 42, so I don't think I can?

gethyn1: oh, i didn't see that

src/web/routes/application/steps/address/select-address/select-address.js
YetAnotherMatt: done, thanks

PR #640, opened 2019-12-03
HTBHF-2675 Moved adding count of matching addresses from nunjucks int…

YetAnotherMatt:
I hadn't realised that Nunjucks would cache the select-address form between posts, so have moved the insertion of the number of matching addresses into javascript from the nunjucks template.

### DIFF
--- a/src/web/routes/application/steps/address/select-address/select-address.js
+++ b/src/web/routes/application/steps/address/select-address/select-address.js
@@ -35,11 +35,16 @@ const resetAddressState = (req) => {
 }
 
 const getAddressDataFromSession = (req, addressId) => {
-  const addresses = req.session.postcodeLookupResults.map(buildAddressOption(addressId))
+  const results = req.session.postcodeLookupResults.map(buildAddressOption(addressId))
+  const addressCount = {
+    value: '',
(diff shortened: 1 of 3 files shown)


src/web/routes/application/steps/address/select-address/select-address.js
```diff
@@ -35,11 +35,16 @@ const resetAddressState = (req) => {
...
 const getAddressDataFromSession = (req, addressId) => {
-  const addresses = req.session.postcodeLookupResults.map(buildAddressOption(addressId))
+  const results = req.session.postcodeLookupResults.map(buildAddressOption(addressId))
+  const addressCount = {
+    value: '',
+    text: req.t('address.numberOfAddressesFound', { count: results.length }),
+    disabled: true,
+    selected: !results.some(addr => addr.selected)
+  }
+  const addresses = results.length === 0 ? [] : [ addressCount ].concat(results)
```
gethyn1: This line could be shortened to:
```
const addresses = results.length ? [] : [addressCount, ...results]
```

gethyn1: .. actually - wouldn't `results` return an empty array by default? So line 45 would just need to be:
```
const addresses = [addressCount, ...results]
```

PR #643, opened 2019-12-06
HTBHF-1942 update test NINO to match V3 claimant service

gethyn1:
Updated `generateEligibleNino` function in test code to match new NINO requirements in claimant service V3 endpoint.

No unit test included as there is not a reliable way of testing against invalid NINOs. If this function does not work then CD pipeline would break.

### DIFF
--- a/src/test/a11y/constants.js
+++ b/src/test/a11y/constants.js
@@ -3,12 +3,17 @@ function randomCharFromChars (chars) {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/test/a11y/constants.js
```diff
@@ -6,9 +6,14 @@ function randomCharFromChars (chars) {
  * Generates a two character string where one character is always `E` to ensure an eligible
```
YetAnotherMatt: Can we update this comment?

PR #645, opened 2019-12-09
HTBHF-1942 determine decision status for failure scenarios

gethyn1:
`getDecisionStatus` consumes `verificationResult` returned from the claimant service. Ultimately this function will return one of four values:

- SUCCESS
- FAIL
- PENDING
- undefined

In a future PR `getDecisionStatus` will be implemented in the `/decision` route handler to determine which decision template to render.

### DIFF
--- a/src/web/routes/application/steps/decision/get-decision-status.js
+++ b/src/web/routes/application/steps/decision/get-decision-status.js
@@ -1,0 +1,26 @@
+const { equals, compose, prop, cond, always } = require('ramda')
+
+const SUCCESS = 'success'
+const FAIL = 'fail'
+const PENDING = 'pending'
+
+const outcomeNotMatched = equals('not_matched')
+const outcomeNotConfirmed = equals('not_confirmed')
+
+const identityOutcomeNotMatched = compose(outcomeNotMatched, prop('identityOutcome'))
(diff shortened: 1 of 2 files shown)


src/web/routes/application/steps/decision/get-decision-status.js
```diff
@@ -0,0 +1,26 @@
...
+const outcomeNotConfirmed = equals('not_confirmed')
+
+const identityOutcomeNotMatched = compose(outcomeNotMatched, prop('identityOutcome'))
+const eligibilityOutcomeNotConfirmed = compose(outcomeNotConfirmed, prop('eligibilityOutcome'))
+
+const isFailure = result => identityOutcomeNotMatched(result) || eligibilityOutcomeNotConfirmed(result)
+
+const getDecisionStatus = cond([
+  [isFailure, always(FAIL)]
+])
```
gethyn1: `cond` facilitates mapping predicates to decision status. When all scenarios are implemented it should hopefully look like (returns `undefined` by default):
```
const getDecisionStatus = cond([
  [isFailure, always(FAIL)],
  [isSuccess, always(SUCCESS)],
  [isPending, always(PENDING)]
])
```

PR #655, opened 2019-12-13
HTBHF-2739 Instant failure page is shown when a claim result's …

dwybourn:
isPregnantOrAtLeast1ChildMatched flag is false.

### DIFF
--- a/src/web/routes/application/steps/decision/get-decision-status.test.js
+++ b/src/web/routes/application/steps/decision/get-decision-status.test.js
@@ -12,7 +12,8 @@ const SUCCESSFUL_RESULT = {
   pregnantChildDOBMatch: 'not_supplied',
   qualifyingBenefits: 'universal_credit',
   identityOutcome: 'matched',
-  eligibilityOutcome: 'confirmed'
+  eligibilityOutcome: 'confirmed',
+  isPregnantOrAtLeast1ChildMatched: 'true'
 }
 
 const PENDING_RESULT = {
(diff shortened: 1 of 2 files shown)


src/web/routes/application/steps/decision/get-decision-status.test.js
```diff
@@ -60,7 +61,22 @@ const ELIGIBILITY_CONFIRMED_RESULT = {
   pregnantChildDOBMatch: 'not_set',
   qualifyingBenefits: 'not_set',
   identityOutcome: 'matched',
-  eligibilityOutcome: 'confirmed'
+  eligibilityOutcome: 'confirmed',
+  isPregnantOrAtLeast1ChildMatched: 'true'
+
+}
+
+const NOT_PREGNANT_AND_NO_CHILDREN_MATCHED_RESULT = {
```
mikeSmith08: This doesn't seem like a very realistic VerificationResult, are we specifically setting all the verification outcomes to `NOT_SET` when the new flag is set? If not, then we should have some verification outcomes set for make this test be more accurate.

dwybourn: I've updated the response to be matched for address line 1 and postcode and set the qualifying benefit to universal credit.

YetAnotherMatt: `not_set` is not a value we're likely to get for our verification outcomes. `not_held` is much more likely for email & phone

PR #656, opened 2019-12-13
NO-JIRA Removed Mac catalina Safari from compatibility tests

YetAnotherMatt:


mikeSmith08:
I missed the acceptance test PR so would have commented on that, but is removing the failing test really the correct way to proceed with this rather than trying to fix it? The test was passing for several builds before this failure appeared.

PR #662, opened 2019-12-17
AFHS-370 Sanitized postcodes are saved to a new variable.

dwybourn:
The new variable sanitizedPostcode is only used by validate as the address lookup code already performs sanitization (converts to upper and strips all whitespace) before calling os places.


src/web/routes/application/steps/address/postcode/validate.js
```diff
@@ -3,13 +3,19 @@ const { check } = require('express-validator')
 const { translateValidationMessage } = require('../../common/translate-validation-message')
 const { UK_POSTCODE_PATTERN, CHANNEL_ISLANDS_AND_IOM_POSTCODE_PATTERN } = require('../constants')
 
+const checkPostcodeIsValid = (_, { req }) => UK_POSTCODE_PATTERN.test(req.body.sanitizedPostcode)
```
mikeSmith08: I don't understand this notation `(_` what does that mean?

dwybourn: It's a variable that we don't care about.

PR #665, opened 2019-12-18
HTBHF-2032 Ensured that the pending decision screen is shown if there is an address mismatch

YetAnotherMatt:


YetAnotherMatt:
@mikeSmith08 This can be merged in the morning. The only change since you saw it last is updating the acceptance tests.

PR #669, opened 2020-01-07
AFHS-543 A selected address is only required when there are address r…

dwybourn:
…esults to select from.

This fixes a bug where clicking on 'Enter address manually' when there are no search results keeps the user on the select address page.

### DIFF
--- a/src/common/predicates.js
+++ b/src/common/predicates.js
@@ -2,6 +2,10 @@ const { compose, isNil, isEmpty, or, not, equals, type } = require('ramda')
 
 const isNilOrEmpty = (string) => or(isNil(string), isEmpty(string))
 
+const isNilOrEmptyOrUndefined = (value) => or(isUndefined(value), isNilOrEmpty(value))
+
+const notIsUndefinedOrNullOrEmpty = compose(not, isNilOrEmptyOrUndefined)
+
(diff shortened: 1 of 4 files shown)


src/common/predicates.js
```diff
@@ -2,6 +2,10 @@ const { compose, isNil, isEmpty, or, not, equals, type } = require('ramda')
 
 const isNilOrEmpty = (string) => or(isNil(string), isEmpty(string))
 
+const isNilOrEmptyOrUndefined = (value) => or(isUndefined(value), isNilOrEmpty(value))
+
+const notIsUndefinedOrNullOrEmpty = compose(not, isNilOrEmptyOrUndefined)
```
YetAnotherMatt: how about `hasNonEmptyValue`?

dwybourn: That could be confusing when passing in a list. To me that would sound like we're checking if the list contains any empty elements.

dwybourn:
> I'm not quite sure why we need to change the validation - wouldn't it be sufficient to avoid submitting the form - i.e. turn the 'enter address manually' button into a link instead, just styled as a button?

I'm not sure how much work that would involve regarding css, and whether or not there are gds/accessibility issues around styling a link as a button. Doesn't seem like it would be an issue but I would like someone with more content/frontend experience to confirm that.

PR #670, opened 2020-01-09
AFHS-571 Use sanitized postcode for validation and request body.

dwybourn:
Updated the manual address validation to validate the sanitised postcode (upper cased and multiple whitespace replaced) instead of the postcode entered by in by the user.

The request body has also been updated to send the sanitised postcode to the claim service.


src/web/routes/application/steps/address/manual-address/test/validation-composition.test.js
```diff
@@ -72,6 +73,17 @@ test('validation middleware errors for invalid postcode field', async (t) => {
   t.end()
 })
 
+test('validation middleware errors for non UK postcode field', async (t) => {
```
dwybourn: This unit test was missing from when the channel islands regex check was added.

src/web/routes/application/steps/address/request-body.js
```diff
@@ -1,9 +1,11 @@
-const { ADDRESS_KEYS } = require('./constants')
-
-const getAddressFieldFromClaim = (claim) => (address, key) => ({ ...address, [key]: claim[key] })
-
 const requestBody = (session) => ({
-  address: ADDRESS_KEYS.reduce(getAddressFieldFromClaim(session.claim), {})
+  address: {
```
dwybourn: Previous code was getting all of the address fields as entered, which meant the non sanitised postcode was being added to the request body.

YetAnotherMatt: Not sure what the answer should be, but should we be using the ADDRESS_KEYS from constants here?

dwybourn: Updated to use the address constants.

PR #672, opened 2020-01-13
AFHS-588 Remove line breaks from success body text as it causes issues…

mikeSmith08:
…with Safari automated tests.

### DIFF
--- a/src/web/server/locales/en/common.json
+++ b/src/web/server/locales/en/common.json
@@ -132,7 +132,7 @@
   },
   "decisionFallbackSuccessful": {
     "title": "Application successful",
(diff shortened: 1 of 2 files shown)


src/web/server/locales/en/common.json
```diff
@@ -132,7 +132,7 @@
   },
   "decisionFallbackSuccessful": {
```
YetAnotherMatt: presumably this will be deleted before the story is complete?

mikeSmith08: Yes, that will be the next PR, all fallback code will be removed.

PR #676, opened 2020-01-24
AFHS-563 double submit issue on button

nhsbsa-dean:


src/web/views/check-answers.njk
```diff
@@ -42,7 +42,8 @@
       text: buttonText,
       href: termsAndConditionsUrl,
       attributes: {
-        id: 'submit-button'
+        id: 'submit-button',
+        preventDoubleClick: true
```
mikeSmith08: Is this required here, it's not submitting to a form?

AmarkumarJagtap: this is added just to make sure that button functionality should not allow double click

YetAnotherMatt: Mike's point is that this button isn't actually submitting a form. I guess if it doesn't do any harm...

nhsbsa-dean: Your right, probably isn't required, I can remove this if need be?

mikeSmith08: My point here is that it would make sense to either do all these extra buttons or none, just doing this one extra button doesn't make too much sense. The important one is the one below in `form.njk`, the others will do no harm but aren't 100% necessary.

nhsbsa-dean: Yeah makes sense @mikeSmith08 I'll revert this page

PR #679, opened 2020-01-27
placed preventdobuleclick at correct level

AmarkumarJagtap:


adeel718:
I think jyothi is currently writing the acceptance tests for this, can we wait for this before merging in please

PR #683, opened 2020-01-30
NO-JIRA Fixed Pa11y tests

YetAnotherMatt:
Although the Pa11y tests themselves were working, the way we combined the Pa11y results with custom tests (testing to make sure the correct page had been opened) meant that we were ignoring the results. This has been fixed, and the version of Pa11y updated to the latest.


src/test/a11y/results.js
```diff
@@ -8,7 +8,7 @@ const handleTestResults = results => {
 
   if (resultsWithIssues.length) {
     console.error('Accessibility issues have been found:')
-    resultsWithIssues.forEach(console.error)
+    console.error(JSON.stringify(resultsWithIssues, null, 2))
```
mikeSmith08: What do the `null` and `2` values mean for this method call?

src/test/a11y/test-suite.js
```diff
@@ -10,14 +10,7 @@ const { URLS, BASE_URL } = require('./paths')
...
 
-const runAndAggregateIssueChecks = (url, result) => (results, checkFn) => {
-  const resultOfCheck = checkFn(url, result)
-  if (resultOfCheck !== null) {
-    return [...results, resultOfCheck]
-  }
-
-  return results
-}
+const notNull = result => result != null
```
mikeSmith08: We have a `notIsNil` method in `predicates.js` that we could use instead of this.

src/test/a11y/test-suite.js
```diff
@@ -34,8 +27,10 @@ const runPa11yTests = async (pages) => {
       console.log('Testing', page.url)
 
       // GET page URL and run a11y
-      const result = await pa11y(page.url, { IGNORE_RULES, headers, timeout: 45000 })
-      result.issues = page.issueChecks.reduce(runAndAggregateIssueChecks(page.url, result), [])
```
mikeSmith08: Can we get rid of the `runAndAggregateIssueChecks` method now?

src/test/a11y/results.js
dwybourn: Should we add a comment that this is for pretty printing? null could be replaced with _ to show that we actually don't care about this parameter.

YetAnotherMatt: we can't replace numm with `_` as that only works when defining a function, not calling it. I have added a comment.

